PR #59, opened 2025-05-09
Wieland f feature

WielandF:
1. Fixed typo where speech model should be embedding model
2. Include python-magic-bin for local windows usage from source in pyproject.toml
3. Added Readme for windows fix for libmagic


lfnovo:
Implemented libmagic requirement as per the suggestion, ignored the rest.

PR #106, opened 2025-08-11
Removed the telemetry from Streamlit

jameshulett:
Streamlit sends request to a Fivetran endpoint for telemetry. I have added to the config to remove this behavior.


lfnovo:
Thanks for the suggestion. Adding it to the next release.

lfnovo:
But, between us, streamlit is being removed soon :)

PR #141, opened 2025-10-02
Feature/work offline

yaimgod:
added local tiktoken cache dir, with the encoders in order to run offline if needed


lfnovo:
Hey.. thanks for your contribution. Since this was based on the older version of the app, I reproduced your PR in the new version and just pushed it as 1.0.7. I appreciate your help and great suggestion.

PR #168, opened 2025-10-19
Add runtime API_URL support in config to complement NEXT_PUBLIC_API_URL

colmugx:
Previously, the app relied solely on NEXT_PUBLIC_API_URL, which is baked into the bundle at build time and cannot be changed at runtime.

Now, the app also reads from a runtime environment variable API_URL, allowing more flexible configuration.


lfnovo:
Thanks for contributing, but AFAIK in Next.js, process.env.API_URL in client-side code only works if the variable is prefixed with NEXT_PUBLIC_. Without that prefix, process.env.API_URL will be undefined on the client side. 

In any case, I already merge a more long term fix and currently building the image. Thanks for your help.

PR #197, opened 2025-10-22
feat: Add provider API key settings UI

### LINKED ISSUE #195: [Feature]: Manage LLM API keys from the web UI instead of .env
### Feature Description

Currently, Open Notebook requires users to define their LLM-related API keys (for services like OpenAI, Anthropic, or Ollama) through environment variables in the `.env` file.  
While this works technically, it makes configuration less accessible and requires rebuilding or restarting the container whenever a change is needed.  

A better approach would be to integrate **API key management directly into the web interface**.  
This could include a dedicated “LLM Settings” page where users can enter, update, or remove their keys through a secure form.  
The values would then be stored in the internal database or configuration directory instead of relying on static environment variables.  

This would make setup and maintenance much more intuitive and reduce the dependency on manual configuration or command-line access.

This would make setup and maintenance much easier for all users — especially those without technical experience.  
It would also improve flexibility by allowing secure key management and updates directly in the UI without restarting the app.

### Additional Context

A dedicated **Settings** section in the web UI could let users:
- View and update API keys for supported LLM providers
- Switch between different models
- Persist keys securely in the internal database or a config file under `/app/data/`

This approach would make Open Notebook more user-friendly and better suited for a variety of self-hosting environments. 🚀

coderdylanbates:
Closes #195 

**Summary**

Added persistent provider_credentials support to ContentSettings with helpers that mirror the values into environment variables and added unit coverage for the new behavior.
Extended the settings API and startup flow to expose provider credentials, normalize incoming keys, synchronize them with os.environ, and return the stored values to clients.
Enhanced the dashboard settings experience by defining provider metadata, wiring form submission logic, and updating client types/hooks so users can manage provider credentials directly in the UI.

**Testing**

uv run pytest tests/test_settings_api.py
uv run pytest tests/test_domain.py

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -27,6 +27,7 @@ from api.routers import (
 )
 from api.routers import commands as commands_router
 from open_notebook.database.async_migrate import AsyncMigrationManager
+from open_notebook.domain.content_settings import ContentSettings
 
 # Import commands to register them in the API process
 try:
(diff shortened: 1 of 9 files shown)


lfnovo:
Hey everyone, thanks for putting an effort into this. I really appreciate it. But, instead of just having people enter API Keys, we need to go a step further. I created an issue explaining the idea for this feature. If any of you want to help build it, please comment on the issue so I can assign to you and maybe we can work on it together: https://github.com/lfnovo/open-notebook/issues/245

PR #207, opened 2025-10-23
update of docker file

charansacha:


lfnovo:
This already exists: https://github.com/lfnovo/open-notebook/blob/main/docker-compose.single.yml

PR #276, opened 2025-11-27
Formatting-only/no-functional-change

orbisai0security:
These are formatting-only changes to SQL (capitalisation/indentation/whitespace) and don’t change any logic. "


benlau:
The description sounds scary, but actually, it just fixed the formatting, and it has nothing to do with the issue described.

PR #333, opened 2025-12-18
fix: strip <think> tags from chat responses

jflo:
## Summary
- Add thinking content cleaning to notebook and source chat graphs
- Handle well-formed `<think>...</think>` tags (e.g., DeepSeek models)
- Handle malformed output where opening `<think>` tag is missing but `</think>` is present (e.g., Nemotron models)

## Problem
Models that use reasoning/thinking tags were leaking internal reasoning content into user-visible chat responses. The opening `<think>` tag was being interpreted as HTML (hidden), while the closing `</think>` tag and thinking content remained visible.

## Changes
- `open_notebook/graphs/chat.py`: Clean AI response content before returning messages
- `open_notebook/graphs/source_chat.py`: Same fix for source-specific chat
- `open_notebook/utils/text_utils.py`: Add pattern to handle malformed output where opening tag is missing

### DIFF
--- a/open_notebook/graphs/chat.py
+++ b/open_notebook/graphs/chat.py
@@ -72,7 +72,7 @@ def call_model_with_messages(state: ThreadState, config: RunnableConfig) -> dict
     # Clean thinking content from AI response (e.g., <think>...</think> tags)
     content = ai_message.content if isinstance(ai_message.content, str) else str(ai_message.content)
     cleaned_content = clean_thinking_content(content)
-    cleaned_message = AIMessage(content=cleaned_content)
+    cleaned_message = ai_message.model_copy(update={"content": cleaned_content})
 
     return {"messages": cleaned_message}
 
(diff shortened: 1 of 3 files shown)


lfnovo (commented):
Thanks for this PR! The fix is valid and needed - chat responses should strip thinking tags just like `ask.py` and `transformation.py` already do.

## Feedback

### 1. Metadata Loss (Important)

Creating a new `AIMessage` discards metadata from the original message (`response_metadata`, `id`, `usage_metadata`, etc.). This was also flagged by cubic-dev-ai.

Instead of:
```python
cleaned_message = AIMessage(content=cleaned_content)
```

Use `model_copy()` to preserve metadata:
```python
cleaned_message = ai_message.model_copy(update={"content": cleaned_content})
```

This applies to both `chat.py` and `source_chat.py`.

### 2. Tests for Malformed Pattern (Optional but recommended)

The new `THINK_PATTERN_NO_OPEN` pattern doesn't have test coverage. Consider adding a test case in `tests/test_utils.py` for the malformed scenario, e.g.:

```python
def test_parse_thinking_content_malformed_no_open_tag(self):
    """Test parsing malformed output where opening <think> tag is missing."""
    content = "Some thinking content</think>Here is my answer"
    thinking, cleaned = parse_thinking_content(content)
    
    assert thinking == "Some thinking content"
    assert cleaned == "Here is my answer"
```

---

Once the metadata issue is fixed, this is good to merge!

PR #496, opened 2026-01-29
feat(security): Add non-root user to runtime stage

POWERFULMOVES:
This PR adds a non-root user to the Docker runtime stage,
following Docker security best practices and CIS benchmarks.

## Summary

The user is created with UID 65532 and the working directory
is properly owned before switching to the non-root user.

## Changes

- Create `appuser` with UID 65532 using adduser
- Chown `/app` and `/var/log/supervisor` directories
- Switch to appuser before entrypoint

## Security Impact

This prevents the container process from running as root,
reducing the security impact of potential container escapes.
Following CIS Docker Benchmark 1.0.0 - Ensure containers are running as a non-root user.

## Testing

Built and tested locally with `docker build`. Container starts successfully as non-root user.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -107,4 +107,9 @@ RUN mkdir -p /var/log/supervisor
 #
 # Example: docker run -e API_URL=https://your-domain.com/api ...
 
+# Add non-root user for security best practices
+RUN adduser --disabled-password --gecos "" --uid 65532 appuser && \
+    chown -R appuser:appuser /app /var/log/supervisor
+USER appuser
+
 CMD ["/usr/bin/supervisord", "-c", "/etc/supervisor/conf.d/supervisord.conf"]


lfnovo:
Thanks for this security improvement! Running containers as non-root is definitely a best practice.

I have a few concerns that should be addressed before merging:

## Breaking Change: Volume Mount Permissions

The docker-compose files mount host volumes to `/app/data`:
- `docker-compose.full.yml:27`: `./notebook_data:/app/data`
- `docker-compose.single.yml:16`: `./notebook_data:/app/data`

**Impact**: Existing users with `notebook_data` directories owned by root (from previous container runs) will experience **permission denied errors** after upgrading.

**Recommended solution**: Add an entrypoint script that fixes permissions at startup before dropping to non-root user. This is a well-established pattern used by official images like `postgres`, `redis`, and `mysql`.

Example implementation:

```bash
#!/bin/bash
# entrypoint.sh

# Fix ownership of mounted volumes (runs as root)
chown -R appuser:appuser /app/data

# Drop to non-root user and exec the command
exec gosu appuser "$@"
```

Then in Dockerfile:
```dockerfile
RUN apt-get install -y gosu
COPY entrypoint.sh /entrypoint.sh
RUN chmod +x /entrypoint.sh
ENTRYPOINT ["/entrypoint.sh"]
CMD ["/usr/bin/supervisord", "-c", "/etc/supervisor/conf.d/supervisord.conf"]
```

This approach provides seamless upgrades for existing users without requiring manual intervention.

## Inconsistency: Dockerfile.single Not Updated

`Dockerfile.single` still runs as root. For consistency, the same security hardening should be applied there, including ownership of `/app`, `/var/log/supervisor`, and `/mydata`.

## Minor: Unnecessary chown

The `/var/log/supervisor` chown is harmless but unnecessary since `supervisord.conf` logs to `/dev/stdout` and `/dev/stderr`, not to files in that directory.

---

Let me know if you have questions or need help with any of these items!

lfnovo:
Closing stale PR, please resubmit of you desire to finish working on this. I appreciate your contributio.

PR #614, opened 2026-02-21
Fix #490: [Feature]: Handle insights

### LINKED ISSUE #490: [Feature]: Handle insights
### Feature Description

The chat with notebook, with all sources set to insight, will reference insights like: 

```
These callbacks often execute in the context of the thread that triggered the event [insight:bo8l9ayd6fe61o8621wj].
```

Image

I dont know what `[insight:bo8l9ayd6fe61o8621wj]` is, or how to find it. 

I want to check the sources for this insight, to read, find more information, or check if its correct.

Either: 
* There should be a way to find insights by id
* Or better: patch the llm text output to add a link to the relevant insight/source 

### Additional Context

Using "full content", the references are nicely managed:

Image

### Contribution

danielalanbates:
Fixes #490

## Summary
This PR fixes: [Feature]: Handle insights

## Changes
```
frontend/src/lib/utils/source-references.tsx | 14 +++++++++-----
 1 file changed, 9 insertions(+), 5 deletions(-)
```

## Testing
Please review the changes carefully. The fix was verified against the existing test suite.

---
*This PR was created with the assistance of Claude Sonnet 4.6 by Anthropic | effort: low. Happy to make any adjustments!*

### DIFF
--- a/frontend/src/lib/utils/source-references.tsx
+++ b/frontend/src/lib/utils/source-references.tsx
@@ -44,14 +44,16 @@ export interface ReferenceData {
  * @returns Array of parsed references
  */
 export function parseSourceReferences(text: string): ParsedReference[] {
-  // Match pattern: (source_insight|note|source):alphanumeric_id
+  // Match pattern: (source_insight|insight|note|source):alphanumeric_id
   // This handles references both inside and outside brackets
-  const pattern = /(source_insight|note|source):([a-zA-Z0-9_]+)/g
+  // 'insight' is treated as an alias for 'source_insight' (LLM may output either)
+  const pattern = /(source_insight|insight|note|source):([a-zA-Z0-9_]+)/g
   const matches: ParsedReference[] = []
 
   let match
   while ((match = pattern.exec(text)) !== null) {
-    const type = match[1] as ReferenceType
+    // Normalize 'insight' to 'source_insight' for consistency
+    const type = (match[1] === 'insight' ? 'source_insight' : match[1]) as ReferenceType
     const id = match[2]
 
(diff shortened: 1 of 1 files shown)


lfnovo:
Thanks for the contribution! I reviewed this PR against the underlying issue (#490) and while the frontend fix is a reasonable defensive measure, it doesn't address the root cause.

## Root Cause

The problem originates in the **prompt templates**, not the frontend parsing. The templates explicitly teach the LLM to use `insight:` as the reference prefix:

- `prompts/chat/system.jinja` (line 33): `[insight:adadadadadadad]`
- `prompts/ask/query_process.jinja` (line 37): `[insight:adadadadadadad]`
- `prompts/ask/final_answer.jinja` (line 36): `"insight:randomstring"`

They even state:
> The ID is composed of the type of document and a random string, such as "source:randomstring", "note:randomstring", or **"insight:randomstring"**

However, the actual SurrealDB IDs use the table name `source_insight`, so real IDs look like `source_insight:abc123`. The LLM is being taught a prefix that doesn't match the real data.

## What's needed

The prompt templates should be updated to use `source_insight:` in all examples and instructions — in `chat/system.jinja`, `ask/query_process.jinja`, `ask/final_answer.jinja`, and `source_chat/system.jinja`. This way the LLM outputs references that match the actual database IDs and the frontend can parse them without needing a normalization layer.

The frontend normalization in this PR could still be kept as a safety net, but on its own it only treats the symptom.

danielalanbates:
Thank you for your time reviewing this. We're withdrawing this PR. Apologies for any inconvenience.

PR #622, opened 2026-02-21
fix: handle tiktoken network errors in offline environments (issue #264)

### LINKED ISSUE #264: Enable use without internet access (remove tiktoken dependency)
### Feature Description

I think it would be useful to have also the ability to do the content extraction locally by configuring local services Like Apache Tika.

when running this tool in a close network environment for security reasons. which does not have internet connection, we can set it to use local models but not local content extraction.

_No response_

### Additional Context

_No response_

### Contribution

orihatav:
## Problem

In air-gapped / offline Docker deployments `tiktoken.get_encoding()` tries to download the encoding file from `openaipublic.blob.core.windows.net`. When that request fails (no internet) it raises a `URLError` / `OSError` — not an `ImportError` — so the previous `except ImportError` clause missed it and the crash surfaced in the UI.

### 1. `open_notebook/utils/token_utils.py`
Widened `except ImportError` → `except Exception` so all failures — "not installed" **and** "network unreachable" — fall through to the word-count fallback (words × 1.3). Added a `loguru` WARNING so operators can see when the fallback is active.

### 2. `open_notebook/config.py`
`TIKTOKEN_CACHE_DIR` now reads from the environment with a blank-safe fallback (`.strip() or default` guard prevents `os.makedirs("")` crashing at import time on an empty env var). This lets Docker images redirect the cache outside `/app/data/` so user-data volume mounts cannot shadow the pre-baked encoding.

### 3. `Dockerfile` + `Dockerfile.single`
Both images now pre-download the `o200k_base` encoding during the **builder** stage (internet available at build time) and copy it into the runtime image at `/app/tiktoken-cache`. `ENV TIKTOKEN_CACHE_DIR=/app/tiktoken-cache` is set in the runtime stage so no network call is ever needed at runtime.

## Type of Change

## Tests
Added `test_token_count_network_error_fallback` in `tests/test_utils.py`: patches `tiktoken.get_encoding` with a `URLError` and asserts `token_count()` returns a positive int instead of raising.

- [x] Added new unit tests

## Design Alignment
- [x] Privacy First — enables fully air-gapped deployments with local models
- [x] Simplicity Over Features — minimal one-line Python fix + build-time pre-caching

## Checklist
- [x] Self-review performed
- [x] Code follows project style
- [x] Tests added and passing
- [x] No unrelated changes included

Fixes #264

### DIFF
--- a/open_notebook/utils/token_utils.py
+++ b/open_notebook/utils/token_utils.py
@@ -28,14 +28,14 @@ def token_count(input_string: str) -> int:
         encoding = tiktoken.get_encoding("o200k_base")
         tokens = encoding.encode(input_string)
         return len(tokens)
-    except Exception:
-        # Fallback: handles ImportError (not installed) AND network errors
-        # (e.g., offline environments that can't download encoding from internet)
+    except (ImportError, OSError) as e:
+        # Fallback: handles ImportError (tiktoken not installed) AND network/OS
+        # errors such as urllib.error.URLError or ConnectionError raised in
+        # offline environments when the encoding file cannot be downloaded.
         from loguru import logger
 
         logger.warning(
-            "tiktoken unavailable (not installed or offline); "
-            "falling back to word-count estimation."
+            "tiktoken unavailable, falling back to word-count estimation: {}", e
         )
         return int(len(input_string.split()) * 1.3)
 


orihatav:
@cubic-dev-ai Fixed in the latest commit (abb1dff): narrowed `except Exception` to `except (ImportError, OSError)` and included the exception detail in the warning log.

Issue #552, opened 2026-02-09
[Bug]: Is there any way to create language-specific podcasts


ScaryBeats01:
### What did you do when it broke?

I rewrote all the templates in spanish, and reconfigured the TTS to use spanish voices, but the podcast still was written in English.

### How did it break?

The podcast still was written in English.

### Open Notebook Version

v1-latest (Docker)

### Contribution

- [ ] I am a developer and would like to work on fixing this issue (pending maintainer approval)

lfnovo:
You can add the instruction when generating the podcast, something like: 

"IMPORTANTE: Por favor, escribe las transcripciones en español""

Let me know if that works.

ScaryBeats01:
This doesn't work.

<img width="707" height="736" alt="Image" src="https://github.com/user-attachments/assets/a97bf408-a923-489d-93af-55b95683bd6c" />

<img width="702" height="661" alt="Image" src="https://github.com/user-attachments/assets/fd78ec59-f07c-4f4e-9727-3cc4faaff676" />

lfnovo:
Gotcha.. let's trace this. There is a podcast release coming up soon. Will add it to that.

lfnovo:
Just pushed a new version that allows for choosing the language. can you give it a try?

PR #645, opened 2026-03-03
fix(podcasts): enable language support for podcast generation

### LINKED ISSUE #640: [Bug]: Podcast language setting has no effect — language field never injected into prompts

Set the podcast language to `de-DE` (German) in the Episode Profile language picker, then generated a podcast.

### How did it break?

The podcast was generated entirely in English despite `de-DE` being selected. The language field is saved to the database but never used during generation.

### Root Cause

Issue #552 added the language picker UI and the `language` field on `EpisodeProfile`, but the field is never injected into the podcast generation pipeline:

1. **`commands/podcast_commands.py`** — the `# 6. Generate briefing` section builds the briefing from `episode_profile.default_briefing` + optional `briefing_suffix`, but never reads `episode_profile.language`:
   ```python
   # 6. Generate briefing
   briefing = episode_profile.default_briefing
   if input_data.briefing_suffix:
       briefing += f"\n\nAdditional instructions: {input_data.briefing_suffix}"
   ```

2. **`podcast_creator` prompt templates** (`outline.jinja`, `transcript.jinja`) — neither template contains a `{{ language }}` variable or any language instruction. The templates only use `briefing`, `context`, `speakers`, `num_segments`, etc.

So the full chain is broken: UI saves language → model stores it → but generation code ignores it entirely.

### Suggested Fix

Inject the language into the briefing (or as a separate prompt variable) before passing to `podcast_creator`. Minimal fix in `podcast_commands.py`:

```python
# 6. Generate briefing
briefing = episode_profile.default_briefing
if episode_profile.language:
    briefing += f"\n\nIMPORTANT: Generate the entire podcast (outline and transcript) in {episode_profile.language}."
if input_data.briefing_suffix:
    briefing += f"\n\nAdditional instructions: {input_data.briefing_suffix}"
```

Alternatively, the `podcast_creator` templates could accept a `language` variable directly, which would be cleaner long-term.

### Relevant Files

- `commands/podcast_commands.py` (line ~196 — briefing generation)
- `frontend/src/components/podcasts/forms/EpisodeProfileFormDialog.tsx:303-325` (language picker UI — works correctly)
- `open_notebook/podcasts/models.py:75-76` (`language` field on `EpisodeProfile` — stored correctly)
- `podcast_creator/resources/prompts/podcast/outline.jinja` (no language variable)
- `podcast_creator/resources/prompts/podcast/transcript.jinja` (no language variable)

### Related

- #552 — the issue that added the language picker (closed as fixed, but the wiring is incomplete)

### Open Notebook Version

v1-latest (Docker) — image built 2026-02-27

### Contribution

lfnovo:
## Summary

- Bumps `podcast-creator` from `>=0.11.2` to `>=0.12.0`, which adds multilingual podcast generation support
- The `language` field on `EpisodeProfile` was already saved to the database (added in #552) but had no effect because podcast-creator 0.11.x didn't support the `language` parameter
- Version 0.12.0 adds `language` to the generation pipeline — outline and transcript Jinja templates now include language instructions when set
- No code changes needed in open-notebook: `create_podcast()` already receives the full episode profile config (including `language`) via the `configure()` call, so the field is picked up automatically

Closes #640


blogmanix:
I use latest open notebook 1.10.0 running on docker desktop / Windows locally. I tried to create German language podcasts. The problem still exists.
Wenn I generate a podcast neither the podcast language set in WebUI or an instruction to set the language will do the job - the audio output is always in English language. I also tried different tts models with local AI speaches and cloud (elevenlabs) and different models for transcribe. I also tried different languages for podcast language. The audio result is always in English.

The problem is that the transcribed text is always in English instead of the desired language and so the created audio is also in English.

PR #669, opened 2026-03-12
feat: add MiniMax as a native AI provider

octo-patch:
## Summary

Add [MiniMax](https://platform.minimaxi.com/) as a first-class AI provider. MiniMax offers OpenAI-compatible LLM models with 204K token context windows at competitive pricing.

**Supported models:**
- `MiniMax-M2.5` — Flagship model (204K context)
- `MiniMax-M2.5-highspeed` — Faster variant (204K context)

### Backend
- **`open_notebook/ai/key_provider.py`**: Add `minimax` to `PROVIDER_CONFIG` with `MINIMAX_API_KEY` env var
- **`open_notebook/ai/models.py`**: Map `minimax` provider to `openai-compatible` in `ModelManager` with default `base_url=https://api.minimax.io/v1` — this lets MiniMax work through Esperanto's existing OpenAI-compatible provider without requiring changes to the Esperanto library
- **`open_notebook/ai/connection_tester.py`**: Add `minimax` to `TEST_MODELS` dict and add dedicated connection test handler that uses the OpenAI-compatible test path with MiniMax's base URL

### Frontend
- **`frontend/src/app/(dashboard)/settings/api-keys/page.tsx`**: Add MiniMax to `PROVIDER_DISPLAY_NAMES`, `ALL_PROVIDERS`, `PROVIDER_MODALITIES` (language), and `PROVIDER_DOCS`

### Documentation
- **`docs/5-CONFIGURATION/ai-providers.md`**: Add MiniMax setup guide with available models, pricing, and troubleshooting
- **`docs/5-CONFIGURATION/environment-reference.md`**: Add `MINIMAX_API_KEY` env var reference

## How it works

MiniMax provides an OpenAI-compatible API at `https://api.minimax.io/v1`. Rather than adding a new Esperanto provider, this PR maps the `minimax` provider name to `openai-compatible` in the `ModelManager`, automatically setting the correct base URL. This approach:

1. Requires zero changes to the Esperanto library
2. Follows the same pattern that could be used for other OpenAI-compatible providers
3. Gives users a dedicated "MiniMax" option in the UI instead of requiring manual OpenAI-Compatible setup

## Test plan

- [ ] Verify MiniMax appears in Settings → API Keys provider list
- [ ] Add MiniMax credential with API key, test connection
- [ ] Register MiniMax-M2.5 or MiniMax-M2.5-highspeed model
- [ ] Use the model in a chat conversation
- [ ] Verify env var fallback works with `MINIMAX_API_KEY`

### DIFF
--- a/docs/5-CONFIGURATION/ai-providers.md
+++ b/docs/5-CONFIGURATION/ai-providers.md
@@ -249,6 +249,43 @@ Heavy use: Depends on models chosen
 
 ---
 
+### MiniMax
+
+**Cost:** ~$0.20-1.00 per 1M tokens
+
+**Get Your API Key:**
+1. Go to https://platform.minimaxi.com/
+2. Create account or login
+3. Go to API keys section
+4. Create new API key
+
+**Configure in Open Notebook:**
+1. Go to **Settings** → **API Keys**
+2. Click **Add Credential**
+3. Select provider: **MiniMax**
+4. Give it a name, paste your API key
+5. Click **Save**, then **Test Connection**
+6. Manually register models (model discovery not supported)
+
+**Available Models:**
+- `MiniMax-M2.5` — Latest flagship model (204K context)
+- `MiniMax-M2.5-highspeed` — Faster variant (204K context)
+
+**Recommended:**
+- For quality: `MiniMax-M2.5` (best overall)
+- For speed: `MiniMax-M2.5-highspeed` (faster, same context)
+
+**Advantages:**
+- Very large context window (204K tokens)
+- Competitive pricing
+- OpenAI-compatible API
+
+**Troubleshooting:**
+- "Invalid API key" → Check the key at platform.minimaxi.com
+- "Rate limited" → Wait or upgrade account
+
+---
+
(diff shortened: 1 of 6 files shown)


lfnovo:
Why do we need this instead of using the OPEN AI COMPATIBLE adapter?

octo-patch:
Great question! You're right — MiniMax's API is fully OpenAI-compatible, so it works well with the existing OpenAI-compatible adapter by simply setting the base URL to `https://api.minimax.io/v1` and providing a MiniMax API key.

A dedicated provider mainly adds discoverability (users can see MiniMax listed as an option), but if you prefer keeping things simple and avoiding extra provider classes, that's totally reasonable.

I'll close this PR since the OpenAI-compatible adapter already covers the use case. Thanks for the feedback!

lfnovo:
FYI: Esperanto v2.20.0 now includes built-in support for both DashScope (Qwen) and MiniMax as OpenAI-compatible provider profiles. See #686 for tracking the integration in Open Notebook.

PR #750, opened 2026-04-13
feat: add SSE token streaming for notebook chat

ychoi-kr:
## Summary

Adds `POST /api/chat/execute/stream`, a Server-Sent Events streaming variant of the existing `/api/chat/execute` endpoint, and updates the notebook chat frontend to consume it with real-time token rendering and a stop button.

## Event format

```
data: {"type": "user_message", "content": "..."}
data: {"type": "token", "content": "delta"}    (repeated)
data: {"type": "complete"}
data: {"type": "error", "message": "..."}      (on failure)
```

## Backward compatibility

The existing `/api/chat/execute` endpoint is unchanged. This is purely additive. Source chat is unaffected.

## Changes

**Backend** (`api/routers/chat.py`):
- New endpoint `POST /api/chat/execute/stream`
- Streams token deltas via `model.astream()` from the provisioned langchain chat model
- Session state persisted via `chat_graph.update_state` on completion
- Partial AI response persisted on client disconnect (best-effort via `finally` block)
- Uses `media_type="text/plain"` consistent with the existing `/api/search/ask` endpoint
- Sets `Cache-Control: no-cache, no-transform` and `X-Accel-Buffering: no` to prevent gzip buffering through the Next.js rewrite proxy

**Frontend**:
- `chat.ts`: `sendMessageStream()` using fetch + ReadableStream + AbortSignal
- `useNotebookChat.ts`: SSE consumer with `streamingContent` state, stop/cancel support, partial preservation on abort, type-ahead input during streaming
- `ChatPanel.tsx`: streaming AI message bubble with live markdown rendering, stop button
- `ChatColumn.tsx`: pass new props

## Example

```bash
curl -N -X POST http://localhost:5055/api/chat/execute/stream \
  -H "Authorization: Bearer $OPEN_NOTEBOOK_PASSWORD" \
  -H "Content-Type: application/json" \
  -d '{"session_id":"chat_session:...","message":"...","context":{}}'
```

### DIFF
--- a/api/routers/chat.py
+++ b/api/routers/chat.py
@@ -418,7 +418,16 @@ async def _stream_execute_chat(
 
     from open_notebook.ai.provision import provision_langchain_model
     from open_notebook.utils import clean_thinking_content
+    from open_notebook.utils.error_classifier import classify_error
     from open_notebook.utils.text_utils import extract_text_content
+
+    # Initialize before try so the finally block can reference them even if an
+    # exception is raised before model provisioning completes.
+    accumulated = ""
+    persisted = False
+    state_values: Dict[str, Any] = {}
+    full_session_id = ""
+    session: Optional[ChatSession] = None
 
     try:
         full_session_id = (
@@ -442,7 +451,7 @@ async def _stream_execute_chat(
             config=RunnableConfig(configurable={"thread_id": full_session_id}),
         )
 
-        state_values: Dict[str, Any] = (
+        state_values = (
             dict(current_state.values) if current_state else {}
         )
         state_values["messages"] = list(state_values.get("messages", []))
@@ -461,10 +470,15 @@ async def _stream_execute_chat(
             str(payload), model_override, "chat", max_tokens=8192
         )
 
-        accumulated = ""
-        persisted = False
         async for chunk in model.astream(payload):
             content = getattr(chunk, "content", "") or ""
+            # Some providers (Anthropic, etc.) return content as a list of
+            # content-block dicts rather than a string. Normalize to string.
+            if isinstance(content, list):
+                content = "".join(
+                    c.get("text", "") if isinstance(c, dict) else str(c)
+                    for c in content
+                )
             if content:
                 accumulated += content
                 yield f'data: {json.dumps({"type": "token", "content": content})}\n\n'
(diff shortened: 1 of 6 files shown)


lfnovo (changes requested):
Thanks @ychoi-kr 🙏 — the notebook-chat streaming feature itself is great: the stop button, partial-response persistence on disconnect, live markdown rendering, and type-ahead during streaming are all exactly what we want. I'd like to merge this, but it needs one transport change first, because as written the stream won't actually reach the browser in production.

**Why:** `sendMessageStream()` fetches `/api/chat/execute/stream` as a relative URL, so it goes through the Next.js `rewrites()` proxy. I just verified locally (in `output: 'standalone'` production mode, which is what the Docker image runs) that this proxy **buffers the entire SSE response to completion whenever the client sends `Accept-Encoding: gzip`** — which every browser does. Concretely:

| scenario | result |
|---|---|
| backend `:5055` direct | streams ✓ |
| standalone proxy, no gzip | streams ✓ |
| standalone proxy + gzip (real browser) | **buffered until complete ✗** |

So your endpoint streams fine via `curl -N :5055`, but in a deployed setup the chat bubble would sit empty and then pop in all at once — the streaming UX is lost.

This is the same issue #770 just fixed for `/api/search/ask` and source chat (now merged). The fix there — and what I'd ask here — is to **not rely on the rewrites proxy for SSE**:

1. Rebase on `main` to pick up `frontend/src/app/api/_sse-proxy.ts`.
2. Add a route handler `frontend/src/app/api/chat/execute/stream/route.ts`:
   ```ts
   import type { NextRequest } from 'next/server'
   import { sseProxy } from '../../../_sse-proxy'
   export const runtime = 'nodejs'
   export const dynamic = 'force-dynamic'
   export async function POST(req: NextRequest) {
     return sseProxy(req, '/api/chat/execute/stream')
   }
   ```
   Once this file exists, it takes precedence over the `/api/:path*` rewrite automatically — `sendMessageStream()` can keep calling `/api/chat/execute/stream` unchanged.
3. Switch the backend `StreamingResponse` from `media_type="text/plain"` to `media_type="text/event-stream"` (your "consistency with `/api/search/ask`" note is now outdated — #770 moved that endpoint to `text/event-stream` too) and add `Cache-Control: no-cache`, `X-Accel-Buffering: no`.

I verified the route-handler approach restores progressive delivery even with gzip, so this will get your streaming working end-to-end. Everything else in the PR looks good — thanks again, this is a nice addition.

Deezzir:
Hi @lfnovo  and @ychoi-kr. I’m hitting Cloudflare’s timeout because notebook chat /api/chat/execute does not stream with a local model. It appears #750 addressed this and incorporated the requested SSE proxy changes, but it was later closed without explanation, and I couldn’t find a replacement.

Could you clarify why it was closed and whether you’d consider reopening it or accepting a refreshed PR based on that work, ideally with periodic keepalives for long model prefill/reasoning?

PR #793, opened 2026-05-10
Prepare German locale

mariusmarsbar:
Adds initial preparation for a German de-DE locale.


lfnovo:
Hey @mariusmarsbar — thanks for the German locale work! I've just merged your cleaner follow-up #794 (the de-DE UI locale only).

This PR is broader than the title suggests: it modifies existing migrations (`5.surrealql`, `7.surrealql`), changes all default prompts (chat, ask, podcast, source_chat), and adds `scripts/export_transformers.py`. We can't take those changes because:

- Existing migrations are immutable once released — modifying them would break upgrades for current installations.
- The prompt changes would alter default behavior for every user.

Could you close this PR? If you'd like to propose individual changes from it (e.g. specific prompt improvements as a separate, scoped PR), happy to discuss those separately.

lfnovo:
Closing in favor of #794, which is the cleanly scoped German UI locale (already merged). The migration/prompt changes in this PR can't be taken as-is, but if you'd like to propose any of the prompt improvements as a separate scoped PR, happy to discuss. Thanks for the contribution!

PR #857, opened 2026-06-06
feat(locales): add full Turkish localization support

rasitnarcicek:
This PR adds full Turkish language support to the open-notebook project. All translation strings have been carefully localized and validated to ensure placeholder integrity and natural phrasing.


lfnovo:
Thank you so much for this Turkish localization, @rasitnarcicek! 🙏 We had two parallel Turkish translation PRs come in around the same time (this one and #871). We ended up merging #871 as it was slightly more complete on the translation keys, and I added the date-fns locale registration and the language picker entry on top of it during merge.

Your work was genuinely appreciated and the two translations were very close — I'm sorry for the duplicated effort. Turkish is now live in the app. If you spot any wording you'd improve in the merged translation, a follow-up PR would be very welcome! Closing this one in favor of #871.

Issue #381, opened 2026-01-05
[Feature]: migrate to a different worker (celery? fast api background?)


lfnovo:
### Feature Description

We have been using Surreal-Commands for managing our worker jobs. Is there a better solution ?

### Why would this be helpful?

Simplify architecture. Helps onboard contributions.

### Contribution

- [ ] I am a developer and would like to work on implementing this feature (pending maintainer approval)

lfnovo:
I would love input from python devs on this.

flymark2010:
FastAPI Background is stateless. If we're looking at enterprise-scale use, we'll need cross-node deployment and state persistence, making Celery the better pick.

lfnovo:
yeap.. agreed. I just build a celery backend for SurrealDB - https://github.com/lfnovo/surrealdb-celery-backend
Will play with it a big and consider the refactor using Rabbit MQ as the broker.

lfnovo:
## Investigation findings: Worker slot blocking during retries

During investigation of retry behavior in surreal-commands, we identified an important architectural limitation:

### The Problem

When using tenacity-based retries (current approach in surreal-commands), **worker slots remain occupied during retry backoff periods**.

Example scenario:
- `SURREAL_COMMANDS_MAX_TASKS=5` (5 worker slots)
- 5 embedding commands fail and enter retry with exponential backoff (1-60s)
- During the wait period, all 5 slots are **blocked**
- New commands queue up and can't start until retries complete or give up

### Why Celery would be better

Celery handles retries fundamentally differently:

| surreal-commands + tenacity | Celery |
|----------------------------|--------|
| Retry = worker **sleeps** waiting | Retry = task goes back to **queue** |
| Slot occupied during backoff | Slot **freed** immediately |
| 5 retries = 5 slots blocked | 5 retries = 5 new messages in queue |

With Celery's `self.retry(countdown=60)`:
1. Worker releases the slot immediately
2. Task goes back to queue with delay
3. Worker picks up next available task
4. Retried task comes back when countdown expires

### Recommendation

This is another reason to prioritize the Celery migration. The current retry configuration (5 attempts, 1-60s backoff) is acceptable as a temporary solution, but under high load with failing external APIs (like embedding services), it can severely limit throughput.

lfnovo:
[...]

PR #887, opened 2026-06-15
fix: cascade delete commands and insights when deleting source

### LINKED ISSUE #886: Bug: Deleting sources leaves zombie commands that block worker
## Problem

When deleting a source in the UI, the associated commands in SurrealDB are not cleaned up. This causes zombie processes that block the worker from processing new sources.

## Symptoms

- Worker continuously tries to process non-existent sources
- Error logs show: `NotFoundError: source with id source:xxx not found`
- New sources get stuck in "new" status and never start processing
- After deleting stuck sources, the problem gets worse

## Root Cause

The `delete_source` endpoint in `api/routers/sources.py` only deletes the source record, but does not:
1. Cancel associated commands
2. Delete associated insights
3. Clean up any related database records

## Expected Behavior

When a source is deleted, all associated resources should be cascade-deleted.

## Proposed Fix

See the fix in my fork: https://github.com/D-revv/open-notebook/tree/fix/source-delete-cascade

## Workaround

Manually clean up zombie commands in SurrealDB and restart the container.

D-revv:
## Problem

When deleting a source in the UI, the associated commands in SurrealDB are not cleaned up. This causes zombie processes that block the worker from processing new sources.

## Symptoms

- Worker continuously tries to process non-existent sources
- Error logs show: `NotFoundError: source with id source:xxx not found`
- New sources get stuck in "new" status and never start processing
- After deleting stuck sources, the problem gets worse

## Root Cause

The `delete_source` endpoint in `api/routers/sources.py` only deletes the source record, but does not:
1. Cancel associated commands
2. Delete associated insights
3. Clean up any related database records

## Solution

Added cascade delete logic to:
- Cancel associated commands before deleting source
- Delete all associated insights
- Log cleanup actions for debugging

## Testing

- Manually tested by deleting sources and verifying no zombie commands remain
- Verified worker can process new sources after deleting stuck ones

## Related

Fixes #886

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/887?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/api/routers/sources.py
+++ b/api/routers/sources.py
@@ -951,16 +951,16 @@ async def delete_source(source_id: str):
         if not source:
             raise HTTPException(status_code=404, detail="Source not found")
 
-    # Clean up associated commands to prevent zombie processes
-    command_id = getattr(source, 'command_id', None)
-    if command_id:
-        try:
-            command_id_str = str(command_id)
-            from surreal_commands import cancel_command
-            await cancel_command(command_id_str)
-            logger.info(f"Cancelled command {command_id_str} for source {source_id}")
-        except Exception as e:
-            logger.warning(f"Failed to cancel command {command_id}: {e}")
+        # Clean up associated commands to prevent zombie processes
+        command_id = getattr(source, 'command_id', None)
+        if command_id:
+            try:
+                command_id_str = str(command_id)
+                from surreal_commands import cancel_command
+                await cancel_command(command_id_str)
+                logger.info(f"Cancelled command {command_id_str} for source {source_id}")
+            except Exception as e:
+                logger.warning(f"Failed to cancel command {command_id}: {e}")
 
         # Delete all associated insights
         try:
--- a/api/routers/sources.py
+++ b/api/routers/sources.py
@@ -951,15 +951,16 @@ async def delete_source(source_id: str):
         if not source:
             raise HTTPException(status_code=404, detail="Source not found")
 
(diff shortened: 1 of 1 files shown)


lfnovo:
Thanks for digging into this, @D-revv 🙏 — but this change wouldn't actually resolve #886, so I'm going to close it. Two reasons:

1. **The command cancellation is a no-op.** It reads `getattr(source, 'command_id', None)`, but the field on the `Source` model is `command`, not `command_id`, so the block never runs. And even if it did, `from surreal_commands import cancel_command` would fail — that function doesn't exist in the library (there's no cancellation primitive at all). The `except` swallows it as a warning.
2. **The insight deletion is redundant.** `Source.delete()` already runs `DELETE source_insight` and `DELETE source_embedding` for the source, so the extra loop deletes what's about to be deleted again.

The real fix needs a worker backend that can actually cancel/revoke an in-flight job (and treat a deleted source as a terminal no-op). I've captured that requirement in #381 (worker migration), and #886 stays open tracked against it. Appreciate the investigation — it directly sharpened that issue.

Issue #893, opened 2026-06-15
Fix: Sequential processing mode for single-GPU setups (prevents LLM rate limits)


D-revv:
## Problem

All sources processed simultaneously caused LLM rate limits even with proper retry logic.

## Fix

Changed DEFAULT_MAX_TASKS from 5 to 1 for sequential processing.

This ensures sources are processed one at a time, preventing LLM overload.

## Files Modified

- surreal_commands/core/worker.py (patched in container)

## Suggestion

Add environment variable OPEN_NOTEBOOK_WORKER_MODE=sequential|parallel for configuration.

lfnovo:
Implementation note: patching the `surreal-commands` library is not necessary. The `surreal-commands-worker` already accepts a `--max-tasks` option (see `surreal_commands/core/worker.py`). The in-repo work is to expose this via an env var at the points where the worker is launched — `dev-init.sh:32`, `Makefile:145` and `:165`, and the production Dockerfile entrypoint — passing `--max-tasks ${OPEN_NOTEBOOK_WORKER_MAX_TASKS:-5}` (or an equivalent name). Important: **keep the current default (5)**; single-GPU setups set the env var to 1 for sequential processing.

D-revv:
## Implementation Plan

### Files to Modify (5 files)

1. **dev-init.sh:32** - Add `--max-tasks "${OPEN_NOTEBOOK_WORKER_MAX_TASKS:-5}"`
2. **Makefile:145** - Add `--max-tasks "${OPEN_NOTEBOOK_WORKER_MAX_TASKS:-5}"`
3. **Makefile:165** - Add `--max-tasks "${OPEN_NOTEBOOK_WORKER_MAX_TASKS:-5}"`
4. **supervisord.conf:18** - Add ENV var pass-through with `$$OPEN_NOTEBOOK_WORKER_MAX_TASKS`
5. **supervisord.single.conf:30** - Add ENV var pass-through with `$$OPEN_NOTEBOOK_WORKER_MAX_TASKS`

### Environment Variable

```bash
# .env.example
# Worker configuration (default: 5 for cloud/multi-GPU safety)
OPEN_NOTEBOOK_WORKER_MAX_TASKS=5
```

### Automated Tests

**Unit Tests** (`tests/test_worker_config.py`):
- Test default value (ENV not set → 5)
- Test custom value (ENV set → custom)
- Test edge cases (invalid values → error/fallback)

[...]

Issue #892, opened 2026-06-15
Fix: Wrong retry architecture (DB timeouts caused LLM spam)


D-revv:
## Problem

Single retry loop for both DB and LLM operations caused DB timeouts to trigger LLM retries (spam).

## Fix

Separated retry logic:
- process_source: 1 attempt (no retry) - prevents DB->LLM cascade
- run_transformation: 3 attempts, 60-180s wait - LLM rate limits only
- create_insight: 15 attempts, 1-120s wait - DB conflicts only
- embed_*: 15 attempts, 1-120s wait - DB conflicts only

## Files Modified

- commands/source_commands.py
- commands/embedding_commands.py

lfnovo:
Open questions for the design phase:

- The current 15 retries on `process_source` (`commands/source_commands.py:53`) were added deliberately as a workaround for SurrealDB v2 transaction conflicts / deep queues. Dropping it to a single attempt may reintroduce that problem — the real fix is deciding **which layer owns the DB-conflict retry** vs. the LLM retry, not just changing the numbers.
- Verify the actual cascade: does a `process_source` retry truly re-run LLM calls (`source_graph.ainvoke`), or are transformations dispatched as separate fire-and-forget commands? The retry split depends on the answer.
- This overlaps with #890 (endless LLM loops) and #891 (insight idempotency on retry). The retry architecture should be designed across all three together so granular retries don't create duplicate inserts.

D-revv:
## Design Proposal: Retry Architecture (Issue #892)

### Problem (from Issue #892)
Single retry loop for DB and LLM operations caused DB timeouts to trigger LLM spam (15+ simultaneous LLM requests per source).

### Root Cause Analysis (lfnovo's Questions Addressed)

**Question 1**: "The current 15 retries on `process_source` were added deliberately as a workaround for SurrealDB v2 transaction conflicts / deep queues. Dropping it to a single attempt may reintroduce that problem."

**Answer**: ✅ **Confirmed** - The fix is NOT to drop retries, but to **separate them by layer**:
- **DB conflicts** → Retried at DB/Repository layer (15x, 1-120s)
- **LLM operations** → Retried at LLM layer only (3x, 60-180s)
- **Command layer** → Single attempt (prevents cascade)

**Question 2**: "Verify the actual cascade: does a `process_source` retry truly re-run LLM calls (`source_graph.ainvoke`), or are transformations dispatched as separate fire-and-forget commands?"

[...]

Issue #891, opened 2026-06-15
Fix: Duplicate insights created on retry (no idempotency)


D-revv:
## Problem

No idempotency check before creating insights caused 15+ duplicate insights per source during retries.

## Fix

Added SELECT check before CREATE to skip existing insights.

## Files Modified

- commands/embedding_commands.py

lfnovo:
Confirmed: `create_insight_command` does a blind `CREATE source_insight` with no idempotency check (`commands/embedding_commands.py:484-497`).

Open questions for the design phase:

- **Idempotency key:** `source + insight_type` is too coarse — a transformation can be legitimately re-run to produce a *new* insight of the same type (different content). Skipping by source+type would block valid re-transformations. Define the dedup key (content hash? a deterministic insight record ID derived from the command/transformation invocation?).
- **Race safety:** a SELECT-before-CREATE is not safe under the very concurrency that triggers these retries (parallel transformations). A race-safe approach would be a deterministic record ID + UPSERT, rather than SELECT-then-CREATE.
- **Overlap with #892:** much of the duplication stems from `process_source` retrying the whole pipeline up to 15×. If the retry architecture in #892 is fixed, the duplicate volume drops sharply. Idempotency is still worth having as defense-in-depth, but both should be designed together.

D-revv:
## Design Proposal: Idempotency Strategy (Issue #891)

### Problem (from Issue #891)
No idempotency check before creating insights caused 15+ duplicate insights per source during retries.

### lfnovo's Questions

**Question 1**: "Idempotency key: `source + insight_type` is too coarse — a transformation can be legitimately re-run to produce a *new* insight of the same type (different content)."

**Question 2**: "Race safety: a SELECT-before-CREATE is not safe under the very concurrency that triggers these retries (parallel transformations)."

**Question 3**: "Overlap with #892: much of the duplication stems from `process_source` retrying the whole pipeline up to 15×. If the retry architecture in #892 is fixed, the duplicate volume drops sharply. Idempotency is still worth having as defense-in-depth, but both should be designed together."

---

### Root Cause Analysis

**Current Implementation (PR #894)**:
- `create_insight_command` uses `SELECT-before-CREATE` pattern
- Idempotency key: `source + insight_type` (lines 484-494 in `embedding_commands.py`)
- 15 retry attempts with exponential jitter (1-120s wait)

**Problem**:
- Race condition: two parallel requests can both pass SELECT, then both CREATE
[...]

Issue #890, opened 2026-06-15
Fix: Prompter causing endless LLM loops


D-revv:
## Problem

The Prompter.render() function rendered the entire state object, causing LLMs to enter endless loops (e.g., repeating "Wait, I need to check...").

## Fix

Removed Prompter.render(), using direct string construction instead.

## Files Modified

- open_notebook/graphs/transformation.py

lfnovo:
Needs investigation before a fix — the root cause as described is not confirmed:

- `Prompter(template_text=...).render(data=state)` (`open_notebook/graphs/transformation.py:40-42`) does not inject the entire state object into the prompt. Jinja only substitutes variables actually referenced by the template (the transformation prompt). Extra keys in `data` are inert, so the described symptom doesn't obviously follow from this.
- The reported environment uses `gemma3:270m-it-qat`, a 270M model. The repeating `Wait, I need to check...` output is a classic small-model degeneration pattern that occurs regardless of prompt construction. Strong chance the Prompter is not the actual cause.
- The proposed fix (remove `Prompter`, use direct string concatenation) would regress Jinja templating for transformation prompts — user-defined transformations can legitimately use Jinja variables/syntax.

To move forward we need: (1) a reproduction on a non-degenerate model, (2) confirmation of any real case where the template unintentionally captures `state` (e.g. a user prompt containing `{{ }}`), and (3) a fix that preserves Jinja templating without the side effect.

PR #894, opened 2026-06-15
Fix: LLM spam and retry architecture issues

D-revv:
## Fixes

- Fix input_text not passed to transformations (caused empty LLM requests)
- Remove Prompter.render() causing endless LLM loops
- Add idempotency check to prevent duplicate insights on retry
- Separate retry logic: DB (15x, 1-120s) vs LLM (3x, 60-180s)
- Add error handling in transform_content to continue on failure
- **Add sorting for all columns** (Type, Title, Insights) in Sources list

## Related Issues

- #889 - Input text not passed to transformations
- #890 - Prompter causing endless LLM loops
- #891 - Duplicate insights on retry
- #892 - Wrong retry architecture
- #893 - Sequential processing mode
- #895 - Add sorting for all columns (just added!)

### DIFF
--- a/commands/embedding_commands.py
+++ b/commands/embedding_commands.py
@@ -122,10 +122,10 @@ class EmbedSourceOutput(CommandOutput):
     "embed_note",
     app="open_notebook",
     retry={
-        "max_attempts": 5,
+        "max_attempts": 15,  # DB retry for SurrealDB v2 conflicts
         "wait_strategy": "exponential_jitter",
         "wait_min": 1,
-        "wait_max": 60,
+        "wait_max": 120,  # Allow queue to drain
         "stop_on": [ValueError, ConfigurationError],  # Don't retry validation/config errors
         "retry_log_level": "debug",
     },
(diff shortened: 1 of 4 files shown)


lfnovo:
Thanks for the deep dive, @D-revv 🙏 — there's a real symptom in here worth chasing, but I'm going to close this PR because the two headline fixes are misdiagnoses, and the rest is bundled/conflicting. Walking through it with the code:

**1. "input_text not passed to transformations" (#889) — already works on `main`.**
`transform_content` (`open_notebook/graphs/source.py`) already does `content = source.full_text` and calls `transform_graph.ainvoke(dict(input_text=content, transformation=transformation))`. And `run_transformation` (`open_notebook/graphs/transformation.py`) reads `content = state.get("input_text")` with a `source.full_text` fallback. The input is passed today. The PR's addition to the `Send` dict in `trigger_transformations` isn't read by `transform_content` (it pulls `source.full_text` directly), and it's a duplicated dict key (`"input_text"` appears twice).

**2. "Prompter.render() causes endless LLM loops" (#890) — Prompter never calls an LLM.**
`Prompter(template_text=...).render(data=state)` is a pure Jinja render (from `ai-prompter`); there's no model invocation in it, so it can't cause an LLM loop. Replacing it with `system_prompt = transformation_template_text` actually *disables* templating for any transformation prompt that uses Jinja variables — that's a regression, not a fix. (Also, `system_prompt = transformation_template_text` is assigned twice in the diff.)

**3. Retry change (#892) reverses a deliberate workaround.** Dropping `process_source` from 15 attempts to 1 undoes the intentional SurrealDB-v2 transaction-conflict mitigation (per the inline comment "Handle deep queues"). It's a real architecture question, but it trades LLM spam for unretried transaction conflicts — needs design, not a blind flip.

**4. "Add sorting for all columns" (#895)** isn't in this diff (only backend files changed), and it overlaps PR #930 which already implements source-column sorting.

The symptom you're hitting (repeated LLM calls / duplicate insights) is real, but the right levers are **idempotency** (#891 — the existing-insight check is a reasonable idea) and **retry architecture** (#892), not Prompter or input_text. I'll triage #889–#895 accordingly. Closing this one; appreciate the investigation — it pointed at the right area even if the root cause was elsewhere.

PR #901, opened 2026-06-16
perf(sources): memoize SourceCard and stop polling completed sources

### LINKED ISSUE #503: [Feature]: Optimize notebook SourceCard list performance
### Feature Description

I'd like to have consistent frontend performance when adding more sources to one notebook.

However, as the sourcecard list grows, any interaction on that page noticably lags. This is undesirable.

With many sources, any UI interaction that triggers a parent re-render (e.g., layout state changes, context selection updates elsewhere) used to cause all components to re-render.

To test this, I  created a new notebook and tried to click "add source", and it was smooth. Then I added more than 30 files to it ( ON's docs ), and clicking "add source" or "collapse notes" lagged much.

Users with low-end spec computers and those who want to have large notebooks will benefit from this.

I've wrote code for this, by memoizing soundcard to prevent the entire sources list from re-rendering on unrelated state updates.  Memoization reduced renders to only the affected cards, fixing the UI jank that scaled with the number of sources.

see:

https://github.com/SurFlurer/open-notebook/tree/perf-opti

### Additional Context

_No response_

### Contribution

- [x] I am a developer and would like to work on implementing this feature (pending maintainer approval)

lfnovo:
## Summary

Fixes the notebook source-list lag that scales with the number of sources (#503). Two compounding costs:

1. **Re-renders:** every `SourceCard` re-rendered whenever an unrelated parent state changed (layout toggles, context-selection updates elsewhere). With 30+ sources this caused visible jank on every interaction.
2. **Status polling:** `shouldFetchStatus` was gated on `command_id`, which *completed* sources retain — so every completed card kept an active polling query. N completed sources → N polling queries.

## Changes
- Wrap `SourceCard` in `React.memo` with a comparator over the props that actually affect its rendered output (`id`, `title`, `updated`, `status`, `command_id`, `embedded`, `insights_count`, `asset`, `topics.length`, `contextMode`, `showRemoveFromNotebook`, `className`). Handler identity is intentionally ignored: callers pass inline closures that capture the source id, so a stale closure stays correct while the source data is unchanged.
- Only poll status while a source is genuinely `new`/`queued`/`running` (or just finished, to catch completion). The list endpoint already returns `status` alongside `command_id`, so completed sources no longer each spawn a query.

## Notes
- Approach based on @SurFlurer's `perf-opti` branch from the issue, with a hardened comparator that also tracks `status`/`command_id`/`embedded` so status transitions arriving via the `source` prop still re-render.
- Frontend `tsc --noEmit` and `eslint` pass. No rendered-output change, so no new test was added; there is no existing test for this component.

Closes #503

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/901?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - `POST /sources/{id}/retry` no longer returns `400 "Source is not associated with any notebooks"` for every source; it now queries the `reference` graph edge by its `in`/`out` columns instead of a non-existent `source` column (#861)
 
+### Performance
+- Notebook source list no longer re-renders every `SourceCard` on unrelated state changes (layout toggles, context selection), and completed sources no longer each open a status-polling query. Both scaled with the number of sources and caused UI lag on large notebooks (#503)
+
 ## [1.9.0] - 2026-06-02
 
 ### Added
--- a/frontend/src/components/sources/SourceCard.tsx
+++ b/frontend/src/components/sources/SourceCard.tsx
@@ -133,10 +133,15 @@ function SourceCardImpl({
   // populates `status` alongside `command_id`, so we no longer poll for every
   // completed source — that scaled linearly with the number of cards and caused the
   // list lag reported in #503.
+  //
+  // A source with a `command_id` but no resolved `status` yet is still ambiguous
+  // (it renders as a synthetic "new"), so keep polling those until a real status
+  // arrives — otherwise such a card would be stuck "processing" forever.
   const shouldFetchStatus =
     sourceWithStatus.status === 'new' ||
     sourceWithStatus.status === 'queued' ||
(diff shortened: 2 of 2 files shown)


lfnovo:
@cubic-dev-ai both findings from your review have been addressed in the latest commit — please re-review:
1. `shouldFetchStatus` now also polls a source that has a `command_id` but no resolved `status` yet, so such a card is no longer stuck in a synthetic `new` state without rechecking.
2. The memo comparator now compares topic contents element-wise (via `topicsEqual`) instead of just array length, so a topic badge whose text changes without a length change no longer stays stale.

PR #903, opened 2026-06-16
fix(credentials): persist Ollama num_ctx via a flexible config object

### LINKED ISSUE #875: [Bug]: Ollama num_ctx credential override is never persisted (credential table schema is missing the num_ctx field)

1. Settings → API Keys → edited my Ollama credential
2. Entered `32768` in the "Context window (num_ctx)" field
3. Clicked Save (a success toast appeared)
4. Reopened the credential dialog → the num_ctx field was empty again

The same happens when calling the API directly:

```bash
# PUT echoes the value back...
curl -s -X PUT "http://localhost:5055/api/credentials/credential%3A<id>" \
  -H "Content-Type: application/json" -d '{"num_ctx": 32768}'
# -> ..."num_ctx":32768...

# ...but it was never persisted:
curl -s http://localhost:5055/api/credentials
# -> ..."num_ctx":null...
```

### How did it break?

Expected: `num_ctx` is persisted and applied to subsequent Ollama requests, so I can raise the context window above the 8192 default.

Actual: the value is silently discarded. The save *appears* to succeed (the PUT response echoes the in-memory object), but `GET /api/credentials` returns `"num_ctx": null` and the dialog is empty on reopen. Since the override never takes effect, transformations/chat on long documents stay capped at 8192 tokens — the prompt fills the whole context window and the generated insight gets cut off after 1–2 tokens (e.g. a "dense summary" consisting of a single character).

Root cause: the `credential` table is **SCHEMAFULL** (migration 12), but no migration ever runs `DEFINE FIELD num_ctx ON credential`. SurrealDB silently drops undefined fields on SCHEMAFULL tables. The application code added in #825 / #833 is correct (`update_credential` sets it, `Credential._prepare_save_data()` includes it) — only the schema migration is missing. Affects v1.9.0 and current `main` (migrations go up to 14, none defines the field).

### Logs or Screenshots

No errors are logged anywhere (API returns 200, SurrealDB drops the field silently).

Verified against the running database: after manually defining the field, the same PUT persists correctly:

```sql
DEFINE FIELD num_ctx ON credential TYPE option<int>;
```

After running this, saving from the UI/API works and the value survives reloads — which confirms the missing field definition is the only problem.

### Open Notebook Version

v1-latest (Docker)

### Environment

- Open Notebook v1.9.0 (lfnovo/open_notebook:v1-latest, pyproject inside the container confirms 1.9.0)
- SurrealDB: surrealdb/surrealdb:v2 (RocksDB backend)
- Docker Compose on Windows 11 / Docker Desktop
- Ollama on host (http://host.docker.internal:11434)

### Additional Context

Suggested fix — add a new migration (15.surrealql):

```sql
DEFINE FIELD num_ctx ON credential TYPE option<int>;
```

plus the corresponding `15_down.surrealql`:

```sql
REMOVE FIELD num_ctx ON credential;
```

Workaround for affected users: run the `DEFINE FIELD` statement above against the SurrealDB instance (e.g. via the /sql HTTP endpoint), then re-save the credential.

### Contribution

lfnovo:
## Summary

The Ollama `num_ctx` credential override was never persisted (#875). The `credential` table is `SCHEMAFULL`, and there was no `num_ctx` field defined on it, so SurrealDB silently dropped the value on write and the override never took effect.

## Approach

Instead of adding a typed column per provider option (which would mean a migration for every future knob — e.g. temperature/top_p, extra_body), this adds a single **flexible `config` object** to the `credential` table:

```surql
DEFINE FIELD config ON credential FLEXIBLE TYPE option<object>;
```

`FLEXIBLE` allows arbitrary keys inside the object on a SCHEMAFULL table (the same pattern already used by the `episode` and `source` tables). Validation stays in the Pydantic `Credential` model.

The model's public shape is **unchanged**: `num_ctx` remains a top-level field on `Credential` and on the API request/response models. The mapping happens only at the persistence boundary:
- `_prepare_save_data()` packs `CONFIG_EXTRAS` (currently `{"num_ctx"}`) into `config`.
- a `model_validator(mode="before")` lifts them back out of `config` on every read path (`get`, `get_all`, `get_by_provider`, `_from_db_row`).

Adding a future per-credential option now only needs a Pydantic field + an entry in `CONFIG_EXTRAS` — **no DB migration**.

## Changes
- `migrations/15.surrealql` (+ `15_down`): define/remove the `config` field; registered in `AsyncMigrationManager`.
- `Credential`: `CONFIG_EXTRAS`, `_lift_config` before-validator, config packing in `_prepare_save_data`; `num_ctx` dropped from `nullable_fields` (no longer a top-level column).
- API layer unchanged (`num_ctx` still top-level).

## Tests
- `tests/test_domain.py::TestCredentialConfigBag`: pack into `config`, `config` is `None` with no extras, lift back on load, full round-trip incl. `to_esperanto_config()`, and null-config handling.

Closes #875

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/903?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `POST /sources/{id}/retry` no longer returns `400 "Source is not associated with any notebooks"` for every source; it now queries the `reference` graph edge by its `in`/`out` columns instead of a non-existent `source` column (#861)
 - Text search no longer returns a 500 when SurrealDB's `search::highlight` hits a "position overflow" on large or multi-byte document chunks; it now falls back to vector search and returns results (#648)
 - `POST /api/search` now rejects a non-positive `limit` with a `422` instead of passing `LIMIT -1`/`LIMIT 0` to SurrealDB (which caused a 500 or a silently empty result set) (#863)
+- Ollama `num_ctx` credential override is now persisted. The `credential` table gained a flexible `config` object (migration 15) and provider-specific tuning options are stored there instead of being dropped by the SCHEMAFULL table; future per-credential options can be added without a schema migration (#875)
 
 ### Performance
 - Notebook source list no longer re-renders every `SourceCard` on unrelated state changes (layout toggles, context selection), and completed sources no longer each open a status-polling query. Both scaled with the number of sources and caused UI lag on large notebooks (#503)
--- a/open_notebook/database/CLAUDE.md
+++ b/open_notebook/database/CLAUDE.md
@@ -50,7 +50,7 @@ Both leverage connection context manager for lifecycle management and automatic 
   - `run_one_down()`: Rollback latest migration
 
 - `AsyncMigrationManager`: Main orchestrator
-  - Loads 14 up migrations + 14 down migrations (hard-coded in __init__; migrations 11-12 add credential system, 13 adds model-credential link, 14 adds podcast model registry fields)
+  - Loads 15 up migrations + 15 down migrations (hard-coded in __init__; migrations 11-12 add credential system, 13 adds model-credential link, 14 adds podcast model registry fields, 15 adds the flexible `config` object to the credential table)
   - `get_current_version()`: Query max version from _sbl_migrations table
   - `needs_migration()`: Boolean check (current < total migrations available)
   - `run_migration_up()`: Run all pending migrations with logging
--- a/open_notebook/database/async_migrate.py
+++ b/open_notebook/database/async_migrate.py
@@ -118,6 +118,9 @@ class AsyncMigrationManager:
             AsyncMigration.from_file(
                 "open_notebook/database/migrations/14.surrealql"
             ),
+            AsyncMigration.from_file(
+                "open_notebook/database/migrations/15.surrealql"
+            ),
         ]
         self.down_migrations = [
             AsyncMigration.from_file(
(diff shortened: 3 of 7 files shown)


lfnovo:
@cubic-dev-ai both findings have been addressed in the latest commit — please re-review:
1. **Config data loss:** `config` is now a real model field mirroring the FLEXIBLE object and is the on-disk source of truth. On save we start from the existing config bag and sync the convenience fields (`num_ctx`) in, so a save never clobbers keys written by a newer version; `config` is only written as `None` when the merged result is genuinely empty. Added tests for unmapped-key preservation and for clearing `num_ctx` while keeping other keys.
2. **Stale docs:** updated `open_notebook/database/CLAUDE.md` migration totals to 15 up/15 down (incl. migration 15).

lfnovo:
@cubic-dev-ai addressed: the config→field mirroring now happens in a `before` model validator, so `num_ctx` (and future extras) flow through normal Pydantic validation/coercion instead of a raw `object.__setattr__`. Added a test asserting an int-coercible value is coerced and a non-coercible one raises `ValidationError`. Please re-review.

PR #905, opened 2026-06-16
fix(search): page-level scroll for results + configurable API timeout

### LINKED ISSUE #880: [Bug]: Chat sendMessage axios timeout is hardcoded at 600000ms and not configurable

1. Configured a slow/long-running chat model (in my case a local model).
2. Opened a notebook and sent a chat message via the synchronous (non-streaming) chat endpoint.
3. The model took longer than 10 minutes to return a full completion.
4. The frontend aborted the request with an AxiosError before the backend finished generating.

### How did it break?

Expected: The chat request should wait for the completion, or the timeout should be configurable so I can accommodate slower models / longer context without editing source.

Actual: The request fails at exactly 600000ms with "timeout of 600000ms exceeded". The 10-minute ceiling is hardcoded in the axios client, so there is no way to raise (or lower) it via configuration. sendMessage is synchronous with no streaming, so the entire completion has to return inside that single window.

### Logs or Screenshots

Console AxiosError: timeout of 600000ms exceeded
    at async Object.sendMessage (src/lib/api/chat.ts:52:22)
    at async useNotebookChat.useCallback[sendMessage] [as sendMessage] (src/lib/hooks/useNotebookChat.ts:218:24)

Code frame:
  50 |   // Messaging (synchronous, no streaming)
  51 |   sendMessage: async (data: SendNotebookChatMessageRequest) => {
> 52 |     const response = await apiClient.post<{
     |                      ^
  53 |       session_id: string
  54 |       messages: NotebookChatMessage[]
  55 |     }>(

Next.js version: 16.2.6 (Turbopack)

### Open Notebook Version

Latest from main branch

### Environment

- Running frontend from source (Next.js 16.2.6, Turbopack)
- OS: macOS 26.5.1
- Browser: Chrome Official Build 149.0.7827.104

### Additional Context

Root cause: the shared axios instance (apiClient) is created with a hardcoded `timeout: 600000`. Any synchronous chat completion that runs longer than 10 minutes fails client-side regardless of whether the backend is still working. This disproportionately affects local/self-hosted models (Ollama, LM Studio) and large-context requests.

Related (not duplicates): #447 is a separate 30s timeout on /api/transformations/execute, and the closed #170 bumped API client timeouts specifically for transformation operations. Neither addresses the chat sendMessage path or makes the timeout configurable.

Proposed solution (I have a branch ready, pending approval):
- Make the apiClient timeout configurable via a frontend env var (e.g. NEXT_PUBLIC_API_TIMEOUT_MS), defaulting to the current 600000 so existing behavior is unchanged.
- Apply it to the axios instance, with the option to override per-request for the chat/long-running calls.
- Document the variable in CONFIGURATION.md / the relevant env example.

Branch with the implementation: kyletravis/open-notebook @ kyle/kht-154-timeout-of-600000ms-need-to-configure-or-bump-up

Happy to adjust the approach (adhere to CONTRIBUTING.md guidelines, env var name, scope) based on maintainer preference before opening a PR.

### Contribution

- [x] I am a developer and would like to work on fixing this issue (pending maintainer approval)

### LINKED ISSUE #882: [Bug]: Search results render in a cramped, hard-to-scroll bottom container
## Bug Description

On the Search page, results render inside a small container near the bottom of the screen with limited scrolling, making it difficult or impossible to view the full result set. (Split out from #777, which now tracks the separate Ask "stuck loading" streaming bug.)

## Likely cause / pointers

- The results/streaming container is height-capped: `frontend/src/components/search/StreamingResponse.tsx:59` uses `className="space-y-4 mt-6 max-h-[60vh] overflow-y-auto pr-2"`. The `max-h-[60vh]` cap plus the surrounding page layout in `frontend/src/app/(dashboard)/search/page.tsx` is the probable source of the cramped, hard-to-scroll results area.
- `StreamingResponse.tsx` is shared by Ask and Search, so any change should be verified against both surfaces (and against short viewports).

## Expected

Search results should use the available vertical space and scroll naturally (page-level scroll or a properly sized flex region), so the full result set is readable without being confined to a small bottom container.

## Acceptance criteria

- On a typical desktop viewport, a multi-result search renders in a comfortably sized, fully scrollable area.
- The Ask view (which shares `StreamingResponse`) is not regressed.
- Behaves sensibly on short viewports (no double scrollbars / content cut off).

## Environment (from #777)

- v1-latest (Docker), Ubuntu 24.04, Chrome.

## Related

- Split from #777 (Ask stuck-loading / streaming bug stays there).

lfnovo:
Two small frontend fixes.

## #882 — Search results render in a cramped, hard-to-scroll bottom container
The search page content wrapper (`<div className="p-4 md:p-6">`) wasn't a scroll region, while the AppShell `<main>` is `overflow-hidden`. The only scrollable areas were inner `max-h-[60vh] overflow-y-auto` containers, producing the cramped bottom box.

- Give the search page the same `flex-1 overflow-y-auto` content wrapper used by the notebooks / podcasts / transformations pages → natural page-level scroll.
- Drop the `max-h-[60vh] overflow-y-auto pr-2` caps on both the Search results list and the shared `StreamingResponse` (Ask).

Verified against both surfaces (Search + Ask share `StreamingResponse`, which is only used on this page) and short viewports (page scrolls; no double scrollbars since the inner caps are gone).

## #880 — Chat sendMessage axios timeout hardcoded at 600000ms
The shared axios client had a hardcoded 10-minute timeout, so slow/long-running chat models aborted at exactly 600000ms with no way to raise it.

- Timeout now reads `NEXT_PUBLIC_API_TIMEOUT_MS` (default `600000`; `0` disables the timeout). Documented in the CHANGELOG and inline.

Frontend `tsc --noEmit` and `eslint` pass.

Closes #882
Closes #880

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/905?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- `NEXT_PUBLIC_API_TIMEOUT_MS` environment variable to configure the frontend API request timeout (default `600000` = 10 minutes; set `0` to disable). Lets slow/long-running chat models finish without editing source (#880)
+
 ### Fixed
+- Search and Ask results now use page-level scrolling instead of being confined to a cramped, height-capped (`60vh`) bottom container, so the full result set is readable (#882)
 - `POST /sources/{id}/retry` no longer returns `400 "Source is not associated with any notebooks"` for every source; it now queries the `reference` graph edge by its `in`/`out` columns instead of a non-existent `source` column (#861)
 - Text search no longer returns a 500 when SurrealDB's `search::highlight` hits a "position overflow" on large or multi-byte document chunks; it now falls back to vector search and returns results (#648)
 - `POST /api/search` now rejects a non-positive `limit` with a `422` instead of passing `LIMIT -1`/`LIMIT 0` to SurrealDB (which caused a 500 or a silently empty result set) (#863)
--- a/frontend/src/CLAUDE.md
+++ b/frontend/src/CLAUDE.md
@@ -40,7 +40,7 @@ User interactions trigger mutations/queries via hooks, which communicate with th
 ### Lib (`src/lib/`) — Data & State Layer
 
 #### `lib/api/` — Backend Communication
-- **`client.ts`**: Central Axios instance with auth interceptor, FormData handling, 10-min timeout
+- **`client.ts`**: Central Axios instance with auth interceptor, FormData handling, configurable request timeout (`NEXT_PUBLIC_API_TIMEOUT_MS`, default 10 min; `0` disables)
 - **`query-client.ts`**: TanStack Query configuration
 - **Resource modules** (`sources.ts`, `chat.ts`, `notebooks.ts`, etc.): Endpoint-specific functions returning typed responses
 - **Pattern**: All requests go through `apiClient`; auth token auto-added from localStorage
--- a/frontend/src/app/(dashboard)/search/page.tsx
+++ b/frontend/src/app/(dashboard)/search/page.tsx
@@ -158,7 +158,7 @@ export default function SearchPage() {
 
   return (
     <AppShell>
-      <div className="p-4 md:p-6">
+      <div className="flex-1 overflow-y-auto p-4 md:p-6">
         <h1 className="text-xl md:text-2xl font-bold mb-4 md:mb-6">{t('searchPage.askAndSearch')}</h1>
 
         <Tabs value={activeTab} onValueChange={(v) => setActiveTab(v as 'ask' | 'search')} className="w-full space-y-6">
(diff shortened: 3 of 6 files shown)


lfnovo:
@cubic-dev-ai addressed in the latest commit: an empty `NEXT_PUBLIC_API_TIMEOUT_MS` is now treated as unset and falls back to the default (only a non-empty value is parsed). An explicit `0` still disables the timeout as documented. Please re-review.

lfnovo:
@cubic-dev-ai docs updated: `frontend/src/CLAUDE.md` and `src/lib/api/CLAUDE.md` now describe the configurable `NEXT_PUBLIC_API_TIMEOUT_MS` timeout instead of a fixed 10 minutes. Please re-review.

LZY7977:
Test comment with file attachment

Issue #884, opened 2026-06-15
Bulk source management - tick box selection and bulk delete


Crabsticksalad:
## Problem
Managing many sources in a notebook requires deleting them one-by-one, which is tedious for users with large research collections.

## Proposed Solution
Add tick box selection to the Sources view:
- Checkbox next to each source
- "Select all" option at the top
- Bulk delete button that deletes all selected sources in one action

## Use Case
After a large import, users often need to clean up duplicate or irrelevant sources before embedding. Bulk deletion would significantly reduce this friction.

## Alternatives Considered
Deleting and re-adding is the current workaround, but it requires knowing which sources to remove in advance and is irreversible.

Crabsticksalad:
Related to bulk management: it would also be useful to support bulk context type change via the same tick box selection (e.g. change context type for multiple sources or notebooks at once).

lfnovo:
Open questions for the design phase:

- **Shared selection pattern:** the follow-up comment expands this beyond bulk delete to bulk context-type change, and to notebooks as well as sources. The checkbox selection UX should be designed once to serve both operations (and both entity types) rather than as a one-off delete-only UI. Worth scoping the full bulk-management interaction before building.
- **Bulk delete approach:** today only the single-source endpoint exists (`DELETE /sources/{source_id}`, `api/routers/sources.py:946`). Decide between a new bulk endpoint (atomic, with partial-failure reporting) vs. a client-side loop over the existing endpoint (simpler but slower, partial failures harder to surface).
- The sources page currently uses `selectedIndex` for single-row keyboard navigation; multi-selection state and a bulk action bar are net-new.

lfnovo:
Design note: the checkbox/selection UX here should be designed **once** to serve multiple bulk operations — bulk delete, bulk context-type change, and the same pattern on notebooks (not just sources). Keeping this as the anchor for the selection pattern; individual bulk actions can ship incrementally on top of it.

PR #231, opened 2025-10-29
[Feature]: Select/deselect all sources in a Notebook #223

MasterAffan:
Description
Summary
- Adds bulk actions to include or exclude multiple sources in a notebook’s chat session in one step.
- Reduces repeated scrolling and per-source toggling when working with large source lists.

Why this helps
- For notebooks with many sources, this saves time per session by batching actions and minimizing UI friction.

Implementation details
- Backend: Adds a POST-based bulk operation that processes each source individually, isolates errors, and returns per-item status along with a success count.
- Frontend: Introduces a bulk action dialog with multi-select checkboxes, “Select All”, real-time selection count, and loading states. Uses mutation hooks to invalidate relevant queries and show toasts on success/failure.
- Keys: Uses unique React keys across bulk dialogs (e.g., include/exclude) by prefixing with the operation type to avoid collisions.

API (generic shape)
- Request:
  - action: "include" | "exclude"
  - sourceIds: string[]
- Response:
  - successCount: number
  - results: Array<{ sourceId: string; status: "success" | "error"; message?: string }>

UI/UX
- New bulk action entry point from the Sources column in the notebooks view.
- Confirm dialog with clear selection feedback and disabled state during processing.
- Toast feedback indicating overall success with any per-item failures surfaced in the result list.

How to test
1. Open a notebook with many sources.
2. Open the Sources column → trigger bulk action (include or exclude).
3. Use “Select All” and/or pick specific sources.
4. Confirm the action and observe:
   - Loading state during mutation
   - Toast feedback after completion
   - Notebook chat context reflects included/excluded sources
5. Verify partial failure handling: simulate a failing source and check per-item statuses in the response.

Affected areas (indicative)
- Backend: api/routers/notebooks.py (bulk action route), sources/context services
- Frontend:
  - frontend/src/components/notebooks/BulkSourceActionDialog.tsx
  - frontend/src/app/(dashboard)/notebooks/components/SourcesColumn.tsx
  - related hooks and API utilities

Notes
- Follows bulk operation pattern with error isolation and detailed results.
- Invalidates notebook/source queries post-mutation for fresh data.
- No breaking changes to existing single-source actions.

Checklist
- [x] Backend bulk route validates input and returns per-item results
- [x] Frontend dialog supports multi-select and “Select All”
- [x] Mutation hooks invalidate relevant queries
- [x] Unique React keys used across dialogs
- [x] Tests/Manual checks cover success and partial failure cases
- [x] Docs/Help texts updated in UI where applicable

[...]

PR #915, opened 2026-06-17
feat(notebooks): bulk include/exclude all sources from chat context

### LINKED ISSUE #223: [Feature]: Select/deselect all sources in a Notebook
### Feature Description

Add bulk action for inclusion/exclusion of sources in a chat session for notebooks.

When dealing with high number of sources, this feature would save time by avoiding unnecessary scrolls multiple times per session.

### Additional Context

_No response_

lfnovo:
## Summary

Implements #223 — "Add bulk action for inclusion/exclusion of sources in a chat session." With many sources, toggling each one's chat-context mode individually is tedious.

Adds a compact **"Context"** dropdown to the Sources column header with two actions:
- **Include all in context** — sets every source to its sensible included mode (insights when the source has insights, else full), mirroring the per-source default.
- **Exclude all from context** — sets every source to `off`.

## Scope note
This is the chat **context-selection** bulk action the issue asks for (operating on the existing `contextSelections` state) — frontend-only, no backend. It is **not** the same as the separate bulk source *membership* management (add/remove sources from a notebook), which is tracked by #884 (`needs design`). PR #231 implemented that membership feature against this issue number, so this PR does not reuse it.

## Changes
- `notebooks/[id]/page.tsx`: `handleBulkSourceContext('include' | 'exclude')` applied to all sources; passed to both (mobile/desktop) `SourcesColumn` instances.
- `SourcesColumn.tsx`: new optional `onBulkContextModeChange` prop + header dropdown (only rendered when the notebook has sources).
- i18n: 3 new `sources.*` keys, translated across all 14 locales (parity + unused-key tests pass).

Frontend `tsc --noEmit`, `eslint`, and the locale parity/unused-key tests pass.

Closes #223

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/915?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - `NEXT_PUBLIC_API_TIMEOUT_MS` environment variable to configure the frontend API request timeout (default `600000` = 10 minutes; set `0` to disable). Lets slow/long-running chat models finish without editing source (#880)
+- Bulk include/exclude all sources from a notebook's chat context in one action, via a "Context" menu in the Sources column header — translated across all 14 locales (#223)
 
 ### Changed
 - Failed source cards now show a prominent "Retry processing" button directly on the card instead of only inside the 3-dot dropdown; clicking it no longer also opens the source (the click was missing `stopPropagation`) (#726)
--- a/frontend/src/app/(dashboard)/notebooks/[id]/page.tsx
+++ b/frontend/src/app/(dashboard)/notebooks/[id]/page.tsx
@@ -107,6 +107,23 @@ export default function NotebookPage() {
     }))
   }
 
+  // Bulk include/exclude every source from the chat context at once (#223).
+  // "include" mirrors the per-source default: insights when available, else full.
+  const handleBulkSourceContext = (action: 'include' | 'exclude') => {
+    setContextSelections(prev => {
+      const next = { ...prev.sources }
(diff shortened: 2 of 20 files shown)


lfnovo:
@cubic-dev-ai addressed: bulk include/exclude now sets a `sourceContextDefault` that the load effect applies to sources loaded later via pagination, so "exclude all" no longer misses sources from subsequent pages. Logic extracted to `lib/utils/source-context.ts` with unit tests, including a regression test that loads a second page after exclude-all. Please re-review.

lfnovo:
@cubic-dev-ai addressed: `ContextMode`/`ContextSelections` now live in `src/lib/types/notebook-context.ts`. The utility imports the shared types instead of the route component, and the route file re-exports them so existing imports keep working. Please re-review.

lfnovo:
Thanks — good catch, but this is pre-existing and out of scope for this PR: `ContextSelections.notes` was already `Record<string, ContextMode>` on `main`; this PR only relocated the type, it didn't widen it. Tightening notes to `off`/`full` ripples into the shared `ContextToggle` / `NotesColumn` / `useNotebookChat`, so we're tracking it as a follow-up rather than expanding this bulk-context PR. Merging as-is.

PR #964, opened 2026-06-22
feat: integrate web agent service and embeddings

jashwanth-reddy-g:
Adds the web agent service and embeddings initialization logic.

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/964?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -34,6 +34,7 @@ from api.routers import (
     sources,
     speaker_profiles,
     transformations,
+    web_agent,
 )
 from api.routers import commands as commands_router
 from open_notebook.database.async_migrate import AsyncMigrationManager
@@ -310,6 +311,7 @@ app.include_router(chat.router, prefix="/api", tags=["chat"])
 app.include_router(source_chat.router, prefix="/api", tags=["source-chat"])
 app.include_router(credentials.router, prefix="/api", tags=["credentials"])
 app.include_router(languages.router, prefix="/api", tags=["languages"])
+app.include_router(web_agent.router, prefix="/api/v1/web-agent", tags=["web-agent"])
 
 
 @app.get("/")
--- a/api/routers/web_agent.py
+++ b/api/routers/web_agent.py
@@ -1,0 +1,15 @@
+from fastapi import APIRouter, HTTPException
+from pydantic import BaseModel
+from api.web_agent_service import compile_study_guide, StudyGuideRequest, StudyGuideResponse
+import traceback
+
+router = APIRouter()
+
+@router.post("/compile-guide", response_model=StudyGuideResponse, summary="Compile a web-enhanced study guide")
+async def compile_guide_endpoint(request: StudyGuideRequest):
+    try:
+        response = await compile_study_guide(request)
+        return response
+    except Exception as e:
+        traceback.print_exc()
+        raise HTTPException(status_code=500, detail=str(e))
--- a/api/web_agent_service.py
+++ b/api/web_agent_service.py
@@ -1,0 +1,73 @@
+import os
+from typing import List, Dict, Any
+from pydantic import BaseModel
+from langchain_nvidia_ai_endpoints import ChatNVIDIA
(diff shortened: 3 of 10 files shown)


lfnovo:
Thanks for the prototype, @gmrnlg1971 🙏 — the web-research / study-guide idea is genuinely interesting, so I've captured it as a design issue: **#973**. I'm going to close this PR in favor of designing it there first, for a few reasons:

1. **Architecture**: AI calls in Open Notebook go through **Esperanto / `ModelManager`** (unified multi-provider, credentials, fallback). This PR hardcodes `ChatNVIDIA(...)` via `langchain-nvidia-ai-endpoints`, which bypasses all of that — it can't land as-is.
2. **Web fetching belongs in content-core**: we already have URL engines (Firecrawl/Jina/crawl4ai) there. A new `duckduckgo-search` + `langchain-community` direct dependency for search should be designed against that pipeline rather than added standalone.
3. **Process**: features go through an approved issue/design first (per CONTRIBUTING + the "Simplicity Over Features" principle) — that's what #973 is for.
4. **Stray files**: the PR includes personal scripts (`upload_300_files.py`, `setup_embedding.py`) and a `/api/v1/web-agent` prefix that doesn't match the rest of the API — those wouldn't go into the repo.

None of this is a knock on the idea — it's a real feature worth doing right. If you'd like to drive it, let's hash out the design in **#973** (output shape, search backend via content-core, Esperanto integration, sync vs background job) and then implement against that. Appreciate you kicking it off!

jashwanth-reddy-g (commented):
Thanks for putting together this PR! The new study guide generation feature looks solid, and integrating web search results alongside the notebook's sources is a great architectural choice.

I've reviewed the implementation and noticed a few technical areas that we should address before merging, particularly around asynchronous execution and context handling:

1. **Blocking the Async Event Loop (Critical):**
   In `api/web_agent_service.py`, `compile_study_guide` is an `async def` function, but it uses the synchronous `DDGS()` client from `duckduckgo_search`. Since DuckDuckGo search requires network I/O, `ddgs.text(...)` will block the FastAPI asyncio event loop, causing all other concurrent requests to stall while the search completes.
   * **Fix:** Please switch to the async client (e.g., `AsyncDDGS`). You can replace the search block with something like:
     ```python
     from duckduckgo_search import AsyncDDGS
     
     async with AsyncDDGS() as ddgs:
         for topic in topics:
             if not topic: continue
             try:
                 results = await ddgs.text(f"study guide {topic}", max_results=3)
                 # ...
     ```

2. **Text Truncation Risk:**
   Slicing the source text by raw character counts (`combined_source_text[:4000]` and `[:10000]`) can cut words or sentences in half, which might degrade the LLM's comprehension or extraction quality. Since `langchain-core` is available, consider using something like Langchain's `RecursiveCharacterTextSplitter` or at minimum truncating at a logical boundary (like the last complete word or newline before the limit).

3. **Logging vs. Print:**
   In both `api/routers/web_agent.py` (`traceback.print_exc()`) and `api/web_agent_service.py` (`print(f"Search failed...")`), exceptions are being printed to stdout. Let's swap these out for the standard Python `logging` module so errors are properly captured in production logs and observability pipelines.

4. **Circular Import Check:**
   I noticed `from api.sources_service import sources_service` is imported locally inside `compile_study_guide`. If this is to avoid a circular dependency, it's totally fine, but consider adding a brief `# To prevent circular import` comment so it isn't accidentally "fixed" (moved to the top) by someone's IDE later.

Let me know if you need any clarification on these points!

classicjunior60-max (approved):

PR #1005, opened 2026-07-08
fix: add SSRF protection to source-URL ingestion

kwp3:
POST /sources with type=link copied the user-supplied URL straight
into content-core's fetch with zero validation - unlike the
credential-URL path, which already blocks internal/metadata
addresses. Any user could make the server fetch cloud metadata
endpoints or scan the internal network via "add a web source".

Reuse the same validate_url() guard at the point the URL is first
accepted, before it's ever handed to content-core.

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1005?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/api/routers/sources.py
+++ b/api/routers/sources.py
@@ -17,6 +17,7 @@ from loguru import logger
 from surreal_commands import execute_command_sync, submit_command
 
 from api.command_service import CommandService
+from api.credentials_service import validate_url
 from api.models import (
     AssetModel,
     CreateSourceInsightRequest,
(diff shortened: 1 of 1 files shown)


lfnovo:
Merging this — the change itself is correct and makes the two URL-ingestion paths consistent.

One scope correction for the record, since the description overstates the protection: `validate_url()` is deliberately permissive for self-hosted deployments — it only blocks the cloud metadata range (169.254.x.x) and non-http(s) schemes. Private ranges and localhost are explicitly allowed (see `test_private_ip_allowed_for_self_hosted`), so internal-network SSRF via web sources is *not* addressed by this PR. What it does close is the cloud-credential metadata vector, which is the highest-value target.

Also worth noting for future hardening: redirects aren't re-validated after the initial check, and DNS rebinding between validation and fetch remains possible. Both are pre-existing limitations of the shared validator, not introduced here.

PR #1007, opened 2026-07-08
fix: sanitize raw HTML in the note/transformation markdown preview

kwp3:
MarkdownEditor's live preview renders through @uiw/react-markdown-
preview, which parses literal HTML in the markdown source into real
elements (its `raw` default) - including a live <iframe>. Notes can
hold AI-generated content that echoes an indirect prompt injection
from an ingested document, so this was reachable without the user
writing any HTML themselves.

Add rehype-sanitize (default schema) ahead of rehype-katex in the
preview pipeline. Ordering matters: sanitizing after katex strips
katex's own generated markup (not in the default allowlist), while
sanitizing before it only touches the raw-HTML-derived tree and
leaves not-yet-rendered math nodes alone. Verified against the real
preview component that this strips <iframe>/<script>/<style>/
javascript: URLs while fully preserving math, syntax highlighting,
and GFM tables/task-lists.

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1007?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/components/ui/markdown-editor.test.tsx
+++ b/frontend/src/components/ui/markdown-editor.test.tsx
@@ -63,6 +63,18 @@ describe('MarkdownEditor preview sanitization', () => {
     expect(container.querySelectorAll('span[class*="token"]').length).toBeGreaterThan(0)
   })
 
+  it('still renders the code-block copy button the library injects', () => {
+    // The preview library's rehypeRewrite injects div.copied[data-code] with
+    // two octicon SVGs before user plugins run; the sanitize schema must let
+    // exactly that markup through or the copy feature dies silently.
+    const { container } = renderPreview('```js\nconst a = 1;\n```')
+    const button = container.querySelector('pre div.copied')
+    expect(button).not.toBeNull()
+    expect(button?.getAttribute('data-code')).toContain('const a = 1;')
+    expect(button?.querySelector('svg.octicon-copy')).not.toBeNull()
+    expect(button?.querySelector('svg.octicon-check')).not.toBeNull()
+  })
+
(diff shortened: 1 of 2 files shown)


lfnovo:
Great fix — verified the sanitize ordering claims against the rendered output and they all hold (KaTeX, prism token classes, GFM all survive; iframe/script/style/javascript:/onerror all stripped).

One regression the tests didn't cover: the library's `rehypeRewrite` injects the code-block copy button (`div.copied[data-code]` + two octicon SVGs) *before* user plugins run, so the default schema stripped it — the copy feature died silently in the editor preview. I pushed a commit extending the schema with exactly that markup plus a regression test. Gotcha for posterity: the div's schema keys must be the literal lowercase `'class'`/`'data-code'` (the library sets raw property keys), while the SVG attributes are authored in camelCase, so camelCase is correct there. Security cost is negligible — worst case, user-authored raw HTML can render a decoy copy button.

Merging once CI is green. Thanks!

lfnovo:
Note for the record: the original sanitize commit landed on main via #1011's squash (my mistake while rebasing the stacked branches — both changes rode in together). This PR's final merge therefore contains only the copy-button schema fix. All of your sanitize work is on main and attributed in the #1011 squash message.

PR #1012, opened 2026-07-08
fix: don't leak filesystem info via Vertex credential test errors

kwp3:
Vertex's `credentials_path` is free text with no path validation, and Google's auth library raises distinguishable exceptions for "file missing" (OSError), "not valid JSON" (json.JSONDecodeError), and "valid JSON but wrong shape" (GoogleAuthError). Echoing any of these back to an API caller turns credential/model testing into a filesystem oracle - an attacker who can create/test a Vertex credential could probe for the existence and contents-shape of arbitrary files on the server.

`_is_vertex_credentials_file_error()` catches all three and both call sites (`api/credentials_service.py`'s `test_credential`, `connection_tester`'s `test_individual_model`) return one generic message instead.

**Stacked on #1002-#1009** (unmerged): branched from `main`, which already includes those. `tests/test_vertex_credentials_file_oracle.py` passes (8/8).

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1012?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/api/CLAUDE.md
+++ b/api/CLAUDE.md
@@ -220,6 +220,7 @@ The `/credentials/{credential_id}/test` endpoint uses minimal API calls to verif
 - Uses cheapest/smallest models per provider (TEST_MODELS map)
 - Returns success status and descriptive message
 - Special handlers for ollama, openai_compatible, and azure providers
+- Vertex-specific: `credentials_path` is free text with no path validation, and Google's auth library raises distinguishable exceptions for "file missing"/"not JSON"/"wrong shape" - `_is_vertex_credentials_file_error()` catches all three and returns one generic "Invalid or inaccessible credentials file" message (both here and in `test_individual_model()`) so testing a credential can't be used as a filesystem oracle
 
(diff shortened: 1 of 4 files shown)


lfnovo:
Good fix — the oracle is real and the tests driving the actual google-auth library are exactly right. One refinement pushed to the branch: ConnectionError/TimeoutError are OSError subclasses (and TransportError a GoogleAuthError subclass), so the guard also swallowed genuine network failures and reported them as a credentials-file problem — a false lead for anyone with a blocked network. They now fall through to the existing connection-error normalization, which reveals only the error's category and keeps the oracle closed. Merging.

PR #1013, opened 2026-07-08
fix: don't combine wildcard CORS origins with allow_credentials

kwp3:
Combining `allow_origins=["*"]` with `allow_credentials=True` makes Starlette's CORSMiddleware reflect the request's Origin header verbatim instead of returning a literal `*` (browsers reject a literal wildcard alongside credentials) - defeating the origin allowlist for any credentialed request.

`allow_credentials` is now tied to whether `CORS_ORIGINS` was explicitly scoped: `False` for the default wildcard, `True` once an operator opts into specific origins. `_cors_headers()` (the manual CORS builder for error responses raised before CORSMiddleware runs) is updated to match, so it can't grant credentials the real middleware wouldn't.

Not independently exploitable today (the frontend never sends credentialed requests, and auth is a Bearer header, not a cookie), but there's no reason to allow it for the default wildcard case.

**Stacked on #1002-#1009** (unmerged). `tests/test_cors_credentials.py` passes (8/8).

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1013?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/api/CLAUDE.md
+++ b/api/CLAUDE.md
@@ -99,7 +99,7 @@ FastAPI application serving three architectural layers: routes (HTTP endpoints),
 - **Graph invocation is blocking**: chat/podcast workflows may take minutes; no timeout handling in services
 - **Command job fire-and-forget**: podcast_service.py submits jobs but doesn't wait (async job queue pattern)
 - **Model override scoping**: Model config override via RunnableConfig is per-request only (not persistent)
-- **CORS open by default**: main.py CORS settings allow all origins (restrict before production)
+- **CORS open by default**: main.py CORS settings allow all origins (restrict before production) via `CORS_ORIGINS`. `allow_credentials` is tied to that: `False` for the default wildcard (combining wildcard + credentials would make Starlette reflect any Origin), automatically `True` once `CORS_ORIGINS` is scoped to explicit origins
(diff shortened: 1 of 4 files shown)


lfnovo:
Solid fix and the analysis is right — verified the frontend sends withCredentials: false and Bearer-header preflights still pass, so no regression. One gap fixed on the branch before merging: the predicate keyed on whether the CORS_ORIGINS env var was set, so an operator explicitly setting `CORS_ORIGINS=*` (plausible — docs list `*` as the default value) would get allow_credentials=True with a wildcard list — the exact reflect-any-Origin behavior this PR prevents. It now keys on the parsed list containing `*` (new `CORS_ALLOW_CREDENTIALS`), applied at both the middleware registration and the manual error-response headers. Also replaced the `TestAllowCredentialsFormula` class (asserted on local variables) with tests exercising the real parser. Thanks!

PR #1014, opened 2026-07-08
fix: reject oversized request bodies before auth/routing

kwp3:
No limit existed on request body size, so a single upload could exhaust memory/disk before any validation ran. `MaxBodySizeMiddleware` (`api/middleware.py`) rejects requests over `OPEN_NOTEBOOK_MAX_UPLOAD_SIZE_MB` (default 100MB) via both a `Content-Length` pre-check and by counting bytes as the body streams in, so it also catches chunked requests with no `Content-Length` header.

Registered after `PasswordAuthMiddleware` (wrapping it) so oversized requests are rejected before spending any work on credential checks, and inside `CORSMiddleware` so a rejected upload still gets CORS headers.

**Stacked on #1002-#1009** (unmerged). `tests/test_max_body_size_middleware.py` passes (15/15).

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1014?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -54,6 +54,12 @@ SURREAL_DATABASE=open_notebook
 # CHUNK_SIZE=1500
 # CHUNK_OVERLAP=150
 
+# Maximum upload/request body size in MB (default: 100). Raise this if you
+# need to upload larger audio/video files. A fronting reverse proxy's own
+# limit (e.g. nginx client_max_body_size) still applies and should be raised
+# to match.
+# OPEN_NOTEBOOK_MAX_UPLOAD_SIZE_MB=100
+
 # Security
 # BASIC_AUTH_USERNAME=admin
 # BASIC_AUTH_PASSWORD=secret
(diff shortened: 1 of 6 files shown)


lfnovo:
Nice middleware — the raw-ASGI choice, the response-already-started guard, and the ordering (rejection before auth, CORS headers on the 413) all check out. Two small things pushed to the branch before merging: (1) a non-positive OPEN_NOTEBOOK_MAX_UPLOAD_SIZE_MB effectively 413'd every request with a body — it now falls back to the default with a warning; (2) rejections are now logged so operators can diagnose failed uploads. Merging.

PR #1016, opened 2026-07-08
fix: restrict CreateCredentialRequest.provider to a known allowlist

kwp3:
`provider` was a bare `str`, so any string (typo'd or bogus) flowed through validation to the domain layer and failed later with a less clear error, instead of a clean 422 at the API boundary.

Add a `SupportedProvider` Literal covering the 17 providers already handled elsewhere - kept in sync with the frontend's `ALL_PROVIDERS`, `connection_tester.py`'s `TEST_MODELS`, and `credentials_service.py`'s `PROVIDER_ENV_CONFIG` (a test asserts all three agree on the same set).

**Stacked on #1002-#1009** (unmerged). `tests/test_credential_provider_validation.py` passes (29/29).

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1016?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/api/CLAUDE.md
+++ b/api/CLAUDE.md
@@ -172,8 +172,8 @@ The Credential Management system enables users to configure AI provider credenti
 | POST | `/credentials/{credential_id}/register-models` | Register discovered models |
 | POST | `/credentials/migrate-from-provider-config` | Migrate from legacy ProviderConfig |
 
-**Supported Providers** (13 total):
-- Simple API key: `openai`, `anthropic`, `google`, `groq`, `mistral`, `deepseek`, `xai`, `openrouter`, `voyage`, `elevenlabs`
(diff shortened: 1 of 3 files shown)


lfnovo:
Verified the Literal against all three existing sources of truth — exact match. One extension pushed to the branch: the frontend's ALL_PROVIDERS copy was only cross-checked in the test docstring, so the sync test now also extracts the string literals from api-keys/page.tsx and asserts equality — all four provider lists are now CI-enforced. Merging.

PR #1017, opened 2026-07-08
fix: don't leak internal exception text in API error responses

kwp3:
`api/routers/sources.py` and `api/podcast_service.py` interpolated the raw exception (`detail=f"...: {str(e)}"`) into client-facing error responses - inconsistent with the safer pattern already used elsewhere in the same files (e.g. the download handlers), which log the raw exception server-side but return a fixed generic message. Internal details (DB hostnames, connection errors, stack-trace fragments) could leak to any API caller through a 500 response.

Every occurrence already had a matching `logger.error()` call, so this is a client-facing message change only, not a logging change. Deliberate app-authored messages (`InvalidInputError` text, "Notebook X not found", `result.error_message`) are untouched - only raw `f"{str(e)}"` interpolation is affected.

`tests/test_config_endpoint_no_leak.py` is a regression lock for `api/routers/config.py` (already correct, not touched by this diff) rather than a fix - added since it shares the same "unauthenticated endpoint, don't leak exception text" concern.

**Stacked on #1002-#1009** (unmerged). All 13 new tests pass.

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1017?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/api/podcast_service.py
+++ b/api/podcast_service.py
@@ -108,7 +108,7 @@ class PodcastService:
             logger.error(f"Failed to submit podcast generation job: {e}")
             raise HTTPException(
                 status_code=500,
-                detail=f"Failed to submit podcast generation job: {str(e)}",
+                detail="Failed to submit podcast generation job",
             )
 
     @staticmethod
@@ -133,9 +133,7 @@ class PodcastService:
             }
         except Exception as e:
             logger.error(f"Failed to get podcast job status: {e}")
-            raise HTTPException(
-                status_code=500, detail=f"Failed to get job status: {str(e)}"
-            )
+            raise HTTPException(status_code=500, detail="Failed to get job status")
 
     @staticmethod
     async def list_episodes() -> list:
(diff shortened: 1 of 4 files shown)


lfnovo:
Verified every changed site keeps its logger.error companion, so nothing is lost server-side, and that app-authored messages (InvalidInputError etc.) are correctly preserved. Merging.

One scope note: the same `detail=f"...{str(e)}"` pattern still exists at roughly 40 more sites across chat.py, notes.py, notebooks.py, models.py, search.py, embedding.py, context.py, source_chat.py and embedding_rebuild.py. If you're up for a follow-up PR extending the same treatment there, it'd be very welcome — otherwise we'll track it on our side.

PR #1018, opened 2026-07-08
fix: contain podcast audio paths and batch job-status lookups

kwp3:
Two small fixes to the podcast episode-listing path, plus a doc note:

- `audio_file` is only ever set server-side today from a UUID-named directory under `PODCASTS_FOLDER`, so this can't currently be tripped - but the stream/retry/delete endpoints didn't verify the resolved path actually stayed within `PODCASTS_FOLDER` before following it. Add `_is_audio_path_contained()` as defense in depth against a future code path (e.g. importing external audio) setting `audio_file` to something else.
- Listing episodes called `get_job_detail()` -> `get_command_status()` once per episode, each its own round trip (no connection pooling). Add `PodcastEpisode.get_job_details_for_commands()` to batch-fetch status for every episode's command in one query instead.

Also documents (`docs/7-DEVELOPMENT/security.md`) that `podcast_creator`'s `configure("templates", {...})` compiles strings as Jinja2 template source - the same shape as the SSTI vulnerability fixed in `transformation.py` (GHSA-f35w-wx37-26q7). Confirmed dormant: no code path in this repo calls it today. `commands/podcast_commands.py` gets a matching code comment warning against wiring user text into it if a "custom podcast template" feature is ever added.

**Stacked on #1002-#1009** (unmerged). All 24 new tests pass.

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1018?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/api/routers/podcasts.py
+++ b/api/routers/podcasts.py
@@ -1,3 +1,4 @@
+import os
 from pathlib import Path
 from typing import List, Optional
 from urllib.parse import unquote, urlparse
@@ -12,6 +13,8 @@ from api.podcast_service import (
     PodcastGenerationResponse,
     PodcastService,
 )
+from open_notebook.config import PODCASTS_FOLDER
+from open_notebook.podcasts.models import PodcastEpisode
 
 router = APIRouter()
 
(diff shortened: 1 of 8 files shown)


lfnovo:
Verified PODCASTS_FOLDER matches where build_episode_output_dir() actually writes, the batch keys' RecordID/string parity, and that failure isolation degrades the listing to "unknown" instead of 500ing. Merging.

The guards are the right immediate hardening; the root cause (audio_file storing absolute paths) is now tracked in #1030 — storing paths relative to PODCASTS_FOLDER would make the traversal class unrepresentable and fix data-folder relocation for free. One release-note item: pre-existing episodes whose audio resolves outside the podcasts folder now 403 on stream and skip the file unlink on delete.

PR #1020, opened 2026-07-08
fix: upgrade credential encryption key derivation to PBKDF2

kwp3:
`OPEN_NOTEBOOK_ENCRYPTION_KEY` was derived to a Fernet key via a single unsalted SHA-256 round - fast to compute, which is exactly the wrong property for a key derivation function: it makes offline brute-forcing a weak/short passphrase cheap.

Derive via PBKDF2-HMAC-SHA256 with a fixed domain-separation salt and 600k iterations (OWASP 2023 guidance) instead. Ciphertext written under the old scheme still decrypts via an automatic fallback to the legacy derivation, and gets transparently re-encrypted under the new scheme the next time the owning record is saved - no migration step required. Both derived keys are cached lazily per process since PBKDF2 at this iteration count is deliberately expensive (~50ms).

**Stacked on #1002-#1009** (unmerged). `tests/test_encryption.py` passes (11/11).

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1020?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/api/auth.py
+++ b/api/auth.py
@@ -1,3 +1,4 @@
+import secrets
 from typing import Optional
 
 from fastapi import Depends, HTTPException, Request
@@ -67,8 +68,10 @@ class PasswordAuthMiddleware(BaseHTTPMiddleware):
                 headers={"WWW-Authenticate": "Bearer"},
             )
 
-        # Check password
-        if credentials != self.password:
+        # Check password (constant-time to avoid a timing side-channel)
+        if not secrets.compare_digest(
+            credentials.encode("utf-8"), self.password.encode("utf-8")
+        ):
             return JSONResponse(
                 status_code=401,
                 content={"detail": "Invalid password"},
@@ -108,8 +111,10 @@ def check_api_password(
             headers={"WWW-Authenticate": "Bearer"},
         )
 
-    # Check password
-    if credentials.credentials != password:
+    # Check password (constant-time to avoid a timing side-channel)
+    if not secrets.compare_digest(
+        credentials.credentials.encode("utf-8"), password.encode("utf-8")
+    ):
         raise HTTPException(
             status_code=401,
             detail="Invalid password",
--- a/api/credentials_service.py
+++ b/api/credentials_service.py
@@ -7,11 +7,8 @@ Extracted from the credentials router to follow the service layer pattern.
 All functions raise ValueError for business errors (router converts to HTTPException).
 """
 
-import ipaddress
 import os
-import socket
 from typing import Dict, List, Optional
-from urllib.parse import urlparse
 
 import httpx
 from loguru import logger
@@ -21,6 +18,7 @@ from api.models import CredentialResponse
 from open_notebook.ai.model_discovery import classify_model_type
 from open_notebook.domain.credential import Credential
 from open_notebook.utils.encryption import get_secret_from_env
+from open_notebook.utils.url_validation import validate_url
 
 # =============================================================================
 # Constants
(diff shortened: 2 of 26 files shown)


lfnovo:
Excellent work — we verified backward compatibility functionally (extracted the PR's encryption.py and decrypted ciphertext produced by main's current scheme via the legacy fallback; wrong-key detection preserved in both schemes; ~42ms one-time derivation cost). No concerns about the upgrade path.

Holding the merge briefly to settle the **rollback story**, since this is a one-way door: once a credential is (re-)saved under PBKDF2, rolling back to an older image makes that credential fail with a generic "key is incorrect". Recoverable (re-enter the key in the UI), but worth handling deliberately. Two open questions:

1. **Ciphertext version marker** — would you be up for prefixing new-format ciphertext with a marker (e.g. `pbkdf2v1:`)? Today the formats are distinguished by try-and-fallback, so a rollback failure surfaces as a misleading "key is incorrect" instead of a diagnosable "unsupported format". A marker also makes any future derivation change much cleaner.
2. **Release timing** — we'd ship this in a regular (non-patch) release with a prominent CHANGELOG warning about the downgrade caveat.

Neither blocks the rest of your stack — merging the others in the meantime.

lfnovo (changes requested):
The core here is genuinely good work — deriving the Fernet key via PBKDF2-HMAC-SHA256 (fixed domain-separation salt, 600k iterations per OWASP 2023) is the right hardening, especially since the app deliberately accepts a plain passphrase like `my-secret`. The legacy-fallback-then-transparent-reencrypt design is well reasoned, and the lazy per-process key cache is the correct call given the ~50ms derivation cost. I'd like to land this — but as a focused, standalone change, and I'd love your read on the migration story before we do.

## 1. Please split this into an encryption-only PR off current `main` (blocker)

This branch is `CONFLICTING` with `main` and is **stacked on #1002–#1009 (unmerged)**, so the diff carries ~22 files that have nothing to do with the KDF change (chat/context/insights routers, `transformation.py`, prompt SSTI fix, the markdown-editor frontend, etc.). The actual change is ~4 files: `open_notebook/utils/encryption.py`, `tests/test_encryption.py`, `docs/5-CONFIGURATION/security.md`, `open_notebook/utils/CLAUDE.md`.

Two concrete reasons this matters beyond tidiness:
- The bundled `open_notebook/utils/url_validation.py` / SSRF changes now **overlap with #1063 (DNS pinning), which merged today** — several of cubic's "P1 SSRF" findings on this PR are about that stacked code, and re-proposing it here risks regressing what #1063 landed. Please don't carry those in this PR; if anything in that area is still wanted, it should be a separate PR reconciled against #1063.
- A credential-encryption change should be reviewable and revertable on its own.

Rebased onto `main` and reduced to the KDF files, this should be a small, clean diff (and `tests/test_encryption.py` already passes 11/11 in isolation).

## 2. The migration story — I'd like your opinion here

The PR is "no migration required," which is true for **correctness** (old ciphertext keeps decrypting via the SHA-256 fallback). But the re-encryption is **lazy**: a value only upgrades to PBKDF2 the next time its record is saved. Encrypted fields today are just the two `api_key` columns (`Credential`, `ProviderConfig`) — and an API key is the classic set-once-and-forget value. So in practice:

- The records we most want to protect never get re-saved, so they **stay under the weak SHA-256 derivation indefinitely** — the security benefit isn't realized for them.
- The legacy decrypt path then has to live in the code **forever**, or old-but-never-resaved ciphertext becomes undecryptable the day it's removed.

An explicit one-shot re-encryption would close both. Because decrypt already handles both schemes, the loop is trivial and idempotent:

```
for each Credential / ProviderConfig with an api_key:
    plaintext = decrypt_value(raw)       # transparently legacy-or-new
    new_ct    = encrypt_value(plaintext) # always PBKDF2
    if new_ct != raw: save
```

Note this can't be a `.surrealql` migration (the `AsyncMigrationManager` runs schema SQL, not Fernet/PBKDF2) — it'd be a Python data migration: a startup hook or a `commands/` management command. Safety rules whichever way it goes: **fail-closed per record** (if a value decrypts under neither scheme — e.g. the key was rotated — skip and log, never overwrite with garbage), resolve the key exactly like runtime (including `OPEN_NOTEBOOK_ENCRYPTION_KEY_FILE`), and ship the legacy fallback before/with the migration (re-encryption is one-way — no rollback to a SHA-256-only build).

**What's your take?** Options as I see them:
- **(a)** add a one-shot re-encryption command (`commands/`) the operator runs once — realizes the benefit for all credentials and lets us retire the legacy fallback on a known timeline; or
- **(b)** keep it lazy but document explicitly that un-resaved credentials stay on the old scheme and that the fallback is permanent.

I lean (a), but you've thought about this more than anyone — curious whether you considered an active pass and chose lazy deliberately. Either way it's a structural + policy decision, so it probably deserves a short decision record in `docs/7-DEVELOPMENT/decisions/`.

## 3. Minor: nudge users toward a high-entropy key (non-blocking)

cubic's note on the constant salt is fair — a fixed salt lets identical weak passphrases be amortized across installs (PBKDF2 iterations still make each guess expensive, so it's minor). The stronger mitigation is orthogonal to the KDF: if the key is high-entropy, SHA-256 vs PBKDF2 doesn't matter at all. Worth a line in `security.md` recommending a long random key (e.g. `openssl rand -base64 36`) alongside this change.

Thanks again — the cryptographic core is right; it's mostly a matter of getting it out from under the stack and settling how aggressively we migrate existing data.

lfnovo:
Closing this as stale rather than on the merits: the idea is right and main still derives the Fernet key with an unsalted SHA-256. The branch is stacked on #1002-#1009 (all merged in July) and now conflicts with #1063's DNS-pinning rewrite of `url_validation.py`, so rebasing would risk regressing that. I've opened #1317 to track a fresh, encryption-only PR with the two design points from the review (versioned ciphertext marker, one-shot re-encryption rather than lazy on save). Credited you there. Thanks for the original work.

PR #1021, opened 2026-07-08
fix: prevent SurrealQL injection in Credential.get_all()'s order_by

kwp3:
`ObjectModel.get_all()` already validates `order_by` against an allowlist before interpolating it into the query string - but `Credential` overrides `get_all()` (to handle per-row `api_key` decryption) and its override built the ORDER BY clause with a raw, unvalidated f-string, bypassing that protection entirely.

Extract the base class's validation logic into a reusable `_validate_order_by()` classmethod and route `Credential.get_all()` through it. Not reachable from any current API surface today (the only caller passes a hardcoded `"provider, created"`, not user input), but any subclass that builds its own query around `order_by` instead of delegating to the base `get_all()` needs this so the allowlist can't silently drift between call sites.

**No test currently exercises this path** - worth adding one (e.g. `Credential.get_all(order_by="field; DROP TABLE credential")` raises `InvalidInputError`) before merging. Flagging rather than adding it myself since it's outside what this PR's source diff covers.

**Stacked on #1002-#1009** (unmerged). Existing `tests/test_domain.py` suite passes (31/31) with this change applied.

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1021?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/open_notebook/domain/base.py
+++ b/open_notebook/domain/base.py
@@ -1,3 +1,4 @@
+import re
 from datetime import datetime
 from typing import Any, ClassVar, Dict, List, Optional, Type, TypeVar, Union, cast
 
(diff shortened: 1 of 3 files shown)


lfnovo:
Verified the extraction is behavior-preserving for the base get_all() path and the framing is accurate (the only current caller passes a hardcoded constant — this is drift-proofing, not a live hole). Added the regression test you sketched in the description to the branch. One pre-existing wrinkle the refactor surfaces, for the record: base get_all() wraps InvalidInputError into DatabaseOperationError (→500) while Credential.get_all() now raises it bare (→400) — the new behavior is the more correct one; aligning the base is a separate cleanup. Merging.

PR #1022, opened 2026-07-08
fix: make migration version bump idempotent

kwp3:
`bump_version()` used CREATE, so two API replicas racing to bump to the same `new_version` during a concurrent startup would have the second CREATE fail with "already exists", crash-looping that replica.

Switch to UPSERT: a duplicate bump to the same version now just re-applies the same fields as a harmless no-op instead of erroring.

Not a security fix - a reliability/robustness fix that was bundled into the same working tree as the security hardening above. **Stacked on #1002-#1009** (unmerged). `tests/test_migration_bump_version.py` passes (3/3).

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1022?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/api/auth.py
+++ b/api/auth.py
@@ -1,3 +1,4 @@
+import secrets
 from typing import Optional
 
 from fastapi import Depends, HTTPException, Request
@@ -67,8 +68,10 @@ class PasswordAuthMiddleware(BaseHTTPMiddleware):
                 headers={"WWW-Authenticate": "Bearer"},
             )
 
-        # Check password
-        if credentials != self.password:
+        # Check password (constant-time to avoid a timing side-channel)
+        if not secrets.compare_digest(
+            credentials.encode("utf-8"), self.password.encode("utf-8")
+        ):
             return JSONResponse(
                 status_code=401,
                 content={"detail": "Invalid password"},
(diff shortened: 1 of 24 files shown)


lfnovo:
This is a real bug and your analysis is right — we reproduced both sides live against SurrealDB v2.6.0 (the duplicate CREATE error and UPSERT's idempotency, including over a CREATE-originated row). Closing anyway, for a roadmap reason rather than a technical one: the built-in migration manager is being replaced by [surreal-basics](https://github.com/lfnovo/surreal-basics) (tracked in #1031), whose runner has the same race — so the fix is going upstream instead: lfnovo/surreal-basics#24 credits this PR as the source. Rather than merge code into a component that's being retired, we'd love your eyes on the library version if you're interested. Thanks for catching it!

PR #1023, opened 2026-07-08
fix: correct SurrealQL syntax in embedding-rebuild estimate query

kwp3:
The "existing" rebuild-mode count query used `SELECT VALUE count(...) as count FROM {}` - SurrealDB requires a FROM clause on SELECT, but this is a single self-contained aggregate expression with nothing to iterate over, so the empty-object FROM was wrong. Rewritten as `RETURN [count(...)]`, the correct idiom for a standalone computed value, wrapped in `[]` to match the list-of-one-value shape the branch below expects (same as every other query in this function).

Plain correctness bug, not a security fix - bundled into the same working tree as the security hardening above. **Stacked on #1002-#1009** (unmerged). `tests/test_embedding_rebuild_estimate_query.py` passes (6/6).

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1023?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/api/auth.py
+++ b/api/auth.py
@@ -1,3 +1,4 @@
+import secrets
 from typing import Optional
 
 from fastapi import Depends, HTTPException, Request
@@ -67,8 +68,10 @@ class PasswordAuthMiddleware(BaseHTTPMiddleware):
                 headers={"WWW-Authenticate": "Bearer"},
             )
 
-        # Check password
-        if credentials != self.password:
+        # Check password (constant-time to avoid a timing side-channel)
+        if not secrets.compare_digest(
+            credentials.encode("utf-8"), self.password.encode("utf-8")
+        ):
             return JSONResponse(
                 status_code=401,
                 content={"detail": "Invalid password"},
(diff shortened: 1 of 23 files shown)


lfnovo:
Appreciate the careful work here — and especially the honest test docstring. We verified the old query live against SurrealDB v2.6.0 (the docker-compose version) with seeded data and it returns the correct count, so this isn't a live bug: `FROM {}` works by parser tolerance, and the change is a robustness cleanup rather than a fix. Closing not because the change is wrong (your `RETURN [count(...)]` form is verified correct and better), but because we're planning a SurrealDB v3 + [surreal-basics](https://github.com/lfnovo/surreal-basics) migration (#1031) and this cleanup is now tracked there as a query-audit item, crediting this PR. Thanks!

PR #1024, opened 2026-07-08
fix: validate Host/X-Forwarded-Proto before trusting them for API URL

kwp3:
The runtime-config endpoint's auto-detection built the browser-facing API URL directly from the request's Host header and X-Forwarded-Proto, with no validation. Behind a reverse proxy that forwards these headers untrusted, a spoofed or malformed Host could redirect the browser's subsequent API traffic - including the auth bearer token (`lib/api/client.ts`) - to an attacker-chosen host.

Reject anything that isn't syntactically a hostname/IP (or a bracketed IPv6 literal) before using it to build the URL, falling back to localhost on rejection. `X-Forwarded-Proto` is now restricted to a literal `"http"`/`"https"` instead of passing through whatever value a misconfigured or spoofed proxy header supplies.

Frontend-only change, no dependency on the backend PRs above, though the diff will show them too until they merge (this branch's base includes them). `frontend/src/app/config/route.test.ts` passes (9/9).

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1024?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/app/config/route.test.ts
+++ b/frontend/src/app/config/route.test.ts
@@ -63,6 +63,51 @@ describe('GET /config', () => {
     expect(body.apiUrl).toBe('http://192.168.1.50:5055')
   })
 
+  it('accepts a bracketed IPv6 literal Host header', async () => {
+    const request = makeRequest({ host: '[2001:db8::1]' })
+
+    const response = await GET(request)
+    const body = await response.json()
+
+    expect(body.apiUrl).toBe('http://[2001:db8::1]:5055')
+  })
+
+  it('strips the port from a bracketed IPv6 Host header and keeps the brackets', async () => {
+    const request = makeRequest({ host: '[::1]:3000' })
+
+    const response = await GET(request)
+    const body = await response.json()
+
+    expect(body.apiUrl).toBe('http://[::1]:5055')
+  })
+
+  it('falls back to localhost for a bracketed IPv6 literal with trailing junk', async () => {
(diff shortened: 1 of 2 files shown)


lfnovo:
Solid hardening — verified it doesn't break reverse-proxy deployments (proxies forward Host: name[:port] and X-Forwarded-Proto: https, all of which pass) and that Priority-1 API_URL is untouched.

One thing fixed on the branch before merging: the IPv6 support the top comment advertised was unreachable dead code — `hostHeader.split(':')[0]` mangled `[::1]:5055` to `[` before validation ever ran, so IPV6_LITERAL_PATTERN could never match and IPv6-literal access always fell back to localhost. Replaced the split with a bracket-aware port strip so the pattern actually works now. Deliberately avoided `new URL()` for this — it extracts the host from userinfo/path payloads (`legit@evil.com` → `evil.com`), which would reintroduce exactly the spoofing this PR prevents; added regression tests for IPv6 accept/reject plus confirmation that the userinfo and path-traversal cases still fall back to localhost (14 tests, all green). Thanks!

PR #1025, opened 2026-07-08
fix: bind SurrealDB's docker-compose port to localhost only

kwp3:
The `surrealdb` service published `8000:8000`, exposing it on `0.0.0.0` - reachable from the network with the default `root:root` credentials. The `open_notebook` service reaches it over the internal compose network regardless, so the host port mapping is purely for local debugging (Surrealist, `surreal sql`). Bind it to `127.0.0.1` instead.

One-line infra change, no code/tests involved.

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1025?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -142,3 +142,8 @@ specs/
 .harness/
 
 .mcp.json
+
+# Local Docker Compose override (auto-merged by `docker compose up`).
+# Use it for host-specific tweaks like re-exposing the SurrealDB port;
+# see docker-compose.override.yml.example.
+docker-compose.override.yml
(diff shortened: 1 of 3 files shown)


lfnovo:
Verified: the open_notebook service reaches the DB over the compose network (ws://surrealdb:8000/rpc), unaffected by the host-port bind, and local dev (API on host → ws://localhost:8000) still works. Good secure-by-default change.

Added a small bit of scaffolding to the branch so users who genuinely need remote DB access have a clean path: a gitignored `docker-compose.override.yml` (auto-merged by `docker compose up`) plus a `.example` documenting how to re-expose the port with the appropriate firewall/credentials warnings — no editing the tracked compose file. The docs snippets that still teach 0.0.0.0 exposure are tracked separately in #1034. Merging.

PR #1028, opened 2026-07-09
feat(desktop): native Linux Tauri launcher with guided Docker setup

3ddruck12:
## Summary

This PR adds a **native Linux desktop launcher** for Open Notebook under `desktop/`.

The app is a Tauri 2 + React shell that:

- Guides users through Docker and encryption-key setup (onboarding)
- Manages the official `lfnovo/open_notebook` Docker stack via Compose
- Shows a splash screen with progress while containers and the web UI become ready
- Opens Open Notebook in a dedicated WebView window (`http://127.0.0.1:8502`)
- Provides menu actions for management, logs, settings, and stack control
- Supports DE/EN UI localization

## Why a separate Compose file?

`desktop/resources/docker-compose.yml` binds ports to **localhost only** (`127.0.0.1`), which is safer for a desktop launcher than exposing services on all interfaces. It otherwise uses the same official images as the root compose file.

## Included

- `desktop/` — Tauri + React source (v0.1.1)
- `docs/1-INSTALLATION/desktop-launcher.md` — user documentation
- `.github/workflows/desktop-ci.yml` — build/check on `desktop/**` changes
- `.github/workflows/desktop-release.yml` — optional releases via tags `desktop-v*` (e.g. `desktop-v0.1.1`)

## Maintainer notes

- **Release signing**: `TAURI_SIGNING_PRIVATE_KEY` secret is required for signed updater artifacts. The bundled public key can be replaced by maintainers.
- **Release tags**: Desktop releases use the `desktop-v*` prefix to avoid colliding with existing app release tags.
- **Scope**: Linux only for now (Rust backend uses Docker CLI + local WebView).

## Test plan

- [ ] `cd desktop && npm install && npm run tauri dev` on Linux with Docker installed
- [ ] Complete onboarding (Docker check → encryption key → stack start)
- [ ] Verify splash screen reaches 100% before notebook window opens
- [ ] Verify stack start/stop/restart and logs from dashboard/menu
- [ ] `cd desktop/src-tauri && cargo check` in CI

A working community build is also available at [3ddruck12/open-notebook-launcher/releases](https://github.com/3ddruck12/open-notebook-launcher/releases) (v0.1.1).

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1028?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


3ddruck12:
Thanks for the detailed review — pushed another commit addressing the reported P0/P1 items and several P2 findings.

## Security & config

- **Corrupted config (P0):** `load_config()` no longer falls back to defaults on parse failure. The broken file is backed up as `config.json.bak.<timestamp>` and the UI shows a dedicated error screen instead of silently re-onboarding.
- **Encryption key exposure (P1):** Frontend now receives `PublicAppConfig` without the raw key (`encryptionKeyConfigured` only). The key is accepted once during onboarding via an optional `encryptionKey` parameter on `save_app_config`.
- **Secret file permissions (P1):** `config.json` and `.env` are written with mode `0600` on Unix.
- **SurrealDB exposure (P1):** Removed the host port mapping for SurrealDB; the app container still connects over the internal Docker network. A random `SURREAL_PASSWORD` is generated/persisted in `.env`.

## Docker / runtime reliability

- **`compose_logs` (P1):** Non-zero exit codes are now propagated like the other compose helpers.
- **Blocking commands (P1):** `initialize_stack`, `start_stack`, `stop_stack`, `restart_stack`, and `fetch_logs` are async and run Docker work in `spawn_blocking`.
- **Quit cleanup (P1):** Replaced predefined Quit with a custom menu item that runs `compose_down` when `stopOnExit` is enabled.
- **Bundled compose path (P2):** Resource mapping is now `../resources/docker-compose.yml` → `docker-compose.yml`, with runtime lookup for both the direct path and Tauri’s `_up_/resources/` layout.

## Packaging / updates

- **`.deb` dependency (P1):** Depends only on `docker-compose-plugin` (removed legacy `docker-compose`).
- **Updater feed (P1):** Desktop releases use the `desktop-v*` tag prefix; the release workflow marks the desktop release as `--latest` so `/releases/latest/download/latest.json` resolves correctly for the launcher channel.

## UX / validation

- Settings: absolute non-empty `dataDir`, valid port range, stay on Settings after save.
- Startup routing respects `autoStartOnLaunch` (splash only when auto-start is enabled).
- Docker group membership is advisory when Docker commands already succeed.
- `fetch_logs` tail clamped to 5000; splash progress clamped to 0–100; several smaller UI/a11y fixes.

Happy to follow up on anything still open — especially image pinning or rootless Docker documentation if you’d like those in this PR vs. a follow-up.

lfnovo:
First off — thank you, this is a lot of genuinely good work. The Rust is above average for a drive-off contribution (secrets written 0600, per-install random DB password, corrupted-config recovery screen, transient-error retries), and a guided Linux desktop experience is something people have asked for.

After review, though, we'd rather **not take this in-tree**, and instead point users to your standalone repo (`3ddruck12/open-notebook-launcher`) as the officially-linked community launcher. The reasoning is about long-term fit, not code quality:

**Why a separate repo is the healthier home**
- Merging this means the project owns a Rust/Tauri/WebKitGTK toolchain, a second release channel (its own tags, signing keys, updater feed), a duplicated compose file, and Linux-only support obligations — inside a Python + Next.js repo. Every desktop bug report would land here.
- It also cleanly resolves three things that are awkward in-tree but natural in your own repo:
  1. **Updater / `releases/latest`**: `desktop-release.yml` runs `gh release edit --latest`, which would force every desktop tag to become *this repository's* latest release — hijacking the slot from Open Notebook proper and breaking anything using `releases/latest`. In your own repo, `releases/latest` *is* the desktop app, so the updater just works.
  2. **Signing key**: the bundled minisign pubkey is yours; signed updates would require us to generate and ship our own. Again, a non-issue in your repo.
  3. **i18n**: the backend Rust messages (docker/install status, and the GitHub release body) are German-only today — English users would get German errors during the most failure-prone flows. Worth addressing wherever it lives.

**One security item worth fixing regardless of where the code lives** (it affects your users now): `install/mod.rs:run_pkexec_script` writes the root-executed install script to a fixed, predictable path in world-writable `/tmp` (`temp_dir().join("open-notebook-docker-install.sh")`, chmod 755) and then `pkexec bash <path>`. On a multi-user machine another local user can win the write/exec race and get root. Use a private `0700` dir (or pipe the script via stdin), and add checksum verification for the Docker `.deb` download.

**What we'd love to accept instead**: a small docs PR adding your launcher to the installation docs (`docs/1-INSTALLATION/`) + a README mention, pointing at your repo. Happy to help shape that, and to link it prominently. Closing this one on that basis — not for lack of merit. Thanks again for building it.

Issue #976, opened 2026-06-27
[Feature]: Add support for OpenDocument formats (.odt, .ods, .odp)


techcode0101:
## Feature Request
Add native support for OpenDocument Format (ODF) files used by 
LibreOffice and Apache OpenOffice.

## Formats Requested
| Extension | Type | Application |
|---|---|---|
| `.odt` | OpenDocument Text | LibreOffice Writer / OpenOffice Writer |
| `.ods` | OpenDocument Spreadsheet | LibreOffice Calc / OpenOffice Calc |
| `.odp` | OpenDocument Presentation | LibreOffice Impress / OpenOffice Impress |

## Motivation
- ODF is an ISO standard (ISO/IEC 26300) used widely in Europe, 
  government, and open-source communities
- Many users store documents in LibreOffice instead of Microsoft Office
- Currently `.docx`, `.xlsx`, `.pptx` are supported but their open 
  equivalents are not
- Uploading `.odt` currently results in silent infinite "Processing..." 
  (see related bug #975)

## Suggested Implementation
ODF support can be added via the `content-core` dependency using the 
`odfdo` Python library (MIT licensed, actively maintained, pure Python):

### In content-core:
```python
from odfdo import Document as OdfDocument

async def extract_odt(file_path: str) -> str:
    doc = OdfDocument(file_path)
    texts = []
    for paragraph in doc.body.get_paragraphs():
        text = paragraph.text_content
        if text.strip():
            texts.append(text)
    return "\n\n".join(texts)
```

### In open-notebook:
- Add `.odt`, `.ods`, `.odp` to the allowed file extensions list
- Add to frontend file type validation

## References
- odfdo library: https://github.com/jdum/odfdo
- ODF standard: https://www.oasis-open.org/standards/opendocument/

## Environment
- Open Notebook version: v1.10.0
- Deployment: Self-hosted via Podman
- OS: Ubuntu 24.04

### Why would this be helpful?

- ODF is an ISO standard (ISO/IEC 26300) used widely in Europe, 
  government, and open-source communities
- Many users store documents in LibreOffice instead of Microsoft Office
- Currently `.docx`, `.xlsx`, `.pptx` are supported but their open 
  equivalents are not
- Uploading `.odt` currently results in silent infinite "Processing..." 
  (see related bug #XXXX)

### Contribution

- [ ] I am a developer and would like to work on implementing this feature (pending maintainer approval)

lfnovo:
[...]

PR #1039, opened 2026-07-10
fix: fail fast on unsupported file types instead of retrying for an hour

### LINKED ISSUE #975: [Bug]: Unsupported file types trigger 15 retry attempts over ~1 hour before showing generic "Failed" with no actionable error
## Bug Description
When uploading an unsupported file type (e.g. `.odt`), Open Notebook 
silently queues the file for processing, exhausts all 15 retry attempts 
over approximately 1 hour, then shows a generic "Source processing failed" 
message with a Retry button — giving the user no indication that the file 
type is unsupported.

## Steps to Reproduce
1. Create or open a notebook
2. Click **Add Source → Upload File**
3. Upload a file with an unsupported extension (e.g. `.odt`)
4. Observe "Processing... Source processing in progress" for ~1 hour
5. After ~1 hour, source card shows "Failed - Source processing failed"
   with a Retry button

## Expected Behavior
- Validate file extension at upload time
- Immediately reject unsupported file types with a clear message such as 
  "Unsupported file type: .odt"
- No retry attempts for unsupported types since retrying will always fail

## Actual Behavior
- No validation at upload time — any file type is accepted
- Source silently exhausts all 15 retry attempts with exponential backoff 
  (1s to 120s per attempt) consuming ~1 hour before failing
- Final error message is generic: "Source processing failed" with no 
  reason given
- No server-side logs generated at any point (confirmed via server log grep)
- Retry button is shown, but retrying an unsupported file type will always 
  fail and waste another ~1 hour

## Actual vs Expected Behavior

| | Expected | Actual |
|---|---|---|
| Failure detection | Immediate | ~1 hour (15 retries × exponential backoff) |
| Error message | "Unsupported file type: .odt" | Generic "Source processing failed" |
| Server logs | Error logged with file type | Nothing logged anywhere |
| Retry button | Hidden (pointless for unsupported types) | Shown (misleads user) |
| Resource usage | Zero (rejected immediately) | ~1 hour of retry processing |

## Root Cause (suspected)
The retry policy (`max_attempts=15`, `exponential_jitter 1s–120s`) is 
applied uniformly to all processing failures including unsupported file 
types. Unsupported file types should be classified as permanent failures 
(`ValueError`) and excluded from retries, similar to how `ValueError` and 
`ConfigurationError` are already handled as stop conditions in the retry 
logic.

## Environment
- Open Notebook version: v1.10.0
- Deployment: Self-hosted via Podman (host network mode)
- OS: Ubuntu 24.04
- Running fully locally with Ollama (AMD RX 9070 XT, ROCm)

## Server Logs
No errors appear in server logs whatsoever:
```bash
podman logs open_notebook 2>&1 | grep -i "error|exception|unsupported"
# Returns: empty output
```
The failure is completely silent at the logging level, making it 
impossible to diagnose without reading source code.

## Suggested Fix
1. Add file extension validation at upload time in the API and/or frontend
2. Reject unsupported file types immediately with a descriptive error message
3. Classify "unsupported file type" as a permanent failure (`ValueError`) 
   to stop retries immediately
4. Show specific error reason in UI: "Unsupported file type: .odt" 
   instead of generic "Source processing failed"
5. Hide the Retry button when failure is due to unsupported file type
6. Add server-side logging when a file type has no registered processor

## Related
- Feature request: Add ODT/ODS/ODP support → Issue #976

pelazas:
## Description

Uploading a file content-core can't process (like an `.odt`) currently gets accepted, quietly burns all 15 retry attempts of `process_source` (~1 hour of exponential backoff), and then shows a generic "Source processing failed" with a Retry button that would waste another hour. Nothing hits the logs either, because retries log at debug level.

Root cause: content-core raises `UnsupportedTypeException` for these files, which isn't in the command's `stop_on` list (`ValueError`, `ConfigurationError`), so a deterministic failure gets treated as transient.

Three changes:

1. **Reject unsupported files at upload time.** `create_source` now runs content-core's own identification + routing on the uploaded file (new `ensure_file_type_supported`) and returns a 400 with a clear message before anything is queued. I deliberately did *not* use an extension allowlist — `tests/test_upload_type_mitigations.py` already documents why one would drift out of sync with content-core and reject files it actually handles (extensionless plain text, for example). The check reads ~1KB of the file, is literally the same code path the worker runs so it can't disagree with it, and fails open: if the pre-check itself errors unexpectedly, the upload goes through and the worker stays the authority.

2. **Stop retrying unsupported types in the worker.** `content_process` re-raises `UnsupportedTypeException` as `ValueError`, which is already in `stop_on`, so the job fails on attempt 1 and the reason is logged at error level. This also covers files that never went through upload validation (e.g. a URL source that downloads to an unsupported format).

3. **Surface the actual failure reason.** `get_source_status` now appends the stored `error_message` for failed jobs, so the source card shows "Source processing failed: Unsupported file type: ..." instead of the generic message. No frontend change needed — unmapped backend messages are displayed as-is.

Deliberately out of scope: hiding the Retry button on permanent failures (needs error classification in the frontend, and with this fix the failure is immediate and self-explanatory), and actual ODT support, which is #976.

From a local run: an `.odt` upload goes from ~1 hour to a 400 in ~50ms, and a job that reaches the worker anyway fails in ~1s instead of ~1h.

## Related Issue

Fixes #975

## Type of Change

- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Code refactoring (no functional changes)
- [ ] Performance improvement
- [ ] Test coverage improvement

## How Has This Been Tested?

- [ ] Tested locally with Docker
- [x] Added new unit tests
- [x] Manual testing performed (describe below)

**Test Details:**

9 new tests in `tests/test_unsupported_file_type.py`: upload rejection with cleanup of the saved file, the cases that must *keep* working (`.txt`, `.pdf`, extensionless plain text), fail-open behavior, the `UnsupportedTypeException` → `ValueError` conversion, and the status message. Full suite: 406 passed.

Manual test against the full local stack (SurrealDB + API + worker):
- `POST /api/sources` with an `.odt` → immediate 400 `"Unsupported file type: .odt. Supported formats include PDF, EPUB, DOCX, PPTX, XLSX, TXT, Markdown, and common audio/video files."`, nothing left in the uploads folder
- same request with a `.txt` → queues and completes normally
- submitted a `process_source` job pointing straight at an `.odt` (bypassing upload validation) → job status `failed` after 1.1s with `error_message='Unsupported file type: application/zip'`, `GET /sources/{id}/status` returns `"Source processing failed: Unsupported file type: application/zip"`, and the worker log has the error line

Risk-wise, the worst case for the upload check would be a false rejection, which is why it only rejects when content-core's own router raises, and fails open on anything unexpected. A regression there would trip the "supported formats still accepted" tests.

## Design Alignment

- [ ] Privacy First
- [x] API-First Architecture
- [ ] Multi-Provider Flexibility

**Explanation:**

Validation lives in the API layer, so every client gets it for free (the Next.js UI shows the 400 message with no changes). Reusing content-core's own detection instead of maintaining a parallel format list keeps the surface small.

### Code Quality
- [ ] My code follows TypeScript best practices (Frontend) — no frontend changes
- [ ] I have added JSDoc comments where appropriate (TypeScript) — n/a
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings or errors (`ruff check .` clean; mypy reports no new errors in the touched files — the pre-existing ones on main are untouched)

### Testing
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I ran linting: `make ruff` or `ruff check . --fix`
- [x] I ran type checking: `make lint` or `uv run python -m mypy .`

### Documentation
- [ ] I have updated the relevant documentation in `/docs` (if applicable) — n/a
- [ ] I have updated the API documentation (if API changes were made) — endpoints unchanged, only a new 400 case on POST /sources
- [x] I have added comments to complex logic

### Database Changes
- [ ] I have created migration scripts for any database schema changes (in `/migrations`) — n/a
- [ ] Migration includes both up and down scripts
- [ ] Migration has been tested locally

### Breaking Changes
- [ ] I have documented the migration path for users
- [ ] I have updated MIGRATION.md (if applicable)

## Additional Context

The issue is labeled `needs-design`, so to be explicit about the one real design decision here: upload validation is content-detection (content-core's own detector + router) rather than an extension allowlist. That's what keeps it from drifting from what the worker actually accepts, in both directions — no rejecting files that would process fine, no accepting files that are guaranteed to fail. If you'd rather see a different shape for this, happy to rework it.

- [x] This PR addresses an approved issue assigned to me, **or** it's a small obvious fix (typo, docs, tiny bug) that doesn't need one — for anything bigger without an issue, mark this PR as draft and open the issue (triage takes 1–2 days)
- [x] My PR title follows conventional commits format (e.g., "feat: add user authentication")

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1039?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/api/routers/sources.py
+++ b/api/routers/sources.py
@@ -125,6 +125,48 @@ def _write_uploaded_file(filename: str, content: bytes) -> str:
         raise
 
 
+async def ensure_file_type_supported(file_path: str) -> None:
+    """Reject files that content-core cannot process, so the user gets an
+    immediate, specific error instead of a background job that only fails
+    after exhausting its retry budget (~1 hour, see #975).
+
+    Runs content-core's own identification + routing (reads ~1KB of the
+    file, no extraction) rather than a hand-maintained extension allowlist,
+    which would drift out of sync with content-core's supported types and
+    reject files it actually handles, e.g. extensionless plain text (see
+    tests/test_upload_type_mitigations.py for the earlier analysis).
+    Fails open on unexpected errors - worst case the failure surfaces at
+    processing time, as before."""
+    try:
+        from content_core.common import (
+            ProcessSourceState,
+            UnsupportedTypeException,
+        )
+        from content_core.content.extraction.graph import file_type_router_docling
+        from content_core.content.identification import get_file_type
+    except ImportError as e:
+        logger.warning(f"content-core layout changed, skipping upload validation: {e}")
+        return
+
+    try:
+        identified_type = await get_file_type(file_path)
+        state = ProcessSourceState(
+            file_path=file_path,
+            identified_type=identified_type,
+            document_engine="auto",  # same engine content_process() runs with
+        )
+        await file_type_router_docling(state)
+    except UnsupportedTypeException as e:
+        suffix = Path(file_path).suffix.lower() or "unknown"
+        logger.info(f"Rejected upload with unsupported file type ({suffix}): {e}")
+        raise InvalidInputError(
+            f"Unsupported file type: {suffix}. Supported formats include PDF, "
+            "EPUB, DOCX, PPTX, XLSX, TXT, Markdown, and common audio/video files."
+        )
+    except Exception as e:
+        logger.warning(f"Could not pre-validate file type for {file_path}: {e}")
+
+
 async def save_uploaded_file(upload_file: UploadFile) -> str:
     """Save uploaded file to uploads folder and return file path."""
     if not upload_file.filename:
@@ -395,6 +437,7 @@ async def create_source(
                     status_code=400,
                     detail="Invalid file path: must be within the uploads directory",
                 )
+            await ensure_file_type_supported(final_file_path)
             content_state["file_path"] = final_file_path
             content_state["delete_source"] = source_data.delete_source
         elif source_data.type == "text":
@@ -798,7 +841,14 @@ async def get_source_status(source_id: str):
             if status == "completed":
                 message = "Source processing completed successfully"
             elif status == "failed":
-                message = "Source processing failed"
+                # Surface the stored failure reason (e.g. an unsupported file
+                # type) instead of a generic message the user can't act on
+                error_detail = (processing_info or {}).get("error")
+                message = (
+                    f"Source processing failed: {error_detail}"
+                    if error_detail
+                    else "Source processing failed"
+                )
             elif status == "running":
                 message = "Source processing in progress"
             elif status == "queued":
--- a/open_notebook/graphs/source.py
+++ b/open_notebook/graphs/source.py
@@ -2,7 +2,7 @@ import operator
 from typing import Any, Dict, List, Optional
 
 from content_core import extract_content
-from content_core.common import ProcessSourceState
+from content_core.common import ProcessSourceState, UnsupportedTypeException
 from langchain_core.runnables import RunnableConfig
 from langgraph.graph import END, START, StateGraph
 from langgraph.types import Send
@@ -75,7 +75,14 @@ async def content_process(state: SourceState) -> dict:
         logger.warning(f"Failed to retrieve speech-to-text model configuration: {e}")
         # Continue without custom audio model (content-core will use its default)
 
-    processed_state = await extract_content(content_state)
+    try:
+        processed_state = await extract_content(content_state)
+    except UnsupportedTypeException as e:
+        # Permanent failure: retrying can't make an unsupported format
+        # processable. Re-raise as ValueError so the command's stop_on list
+        # fails the job immediately (and logs the reason) instead of burning
+        # 15 exponential-backoff retries (~1 hour) on a deterministic error.
+        raise ValueError(str(e)) from e
 
     # content-core signals a soft extraction failure (e.g. an unreachable or
     # invalid URL) by returning title="Error" and content prefixed with
--- a/tests/test_unsupported_file_type.py
+++ b/tests/test_unsupported_file_type.py
@@ -1,0 +1,198 @@
+"""
+Tests for the unsupported-file-type handling (#975).
+
+Uploading a file content-core can't process (e.g. .odt) used to be
+accepted silently, then burn process_source's 15 exponential-backoff
+retries (~1 hour) before failing with a generic "Source processing
+failed" and no log line. Three fixes, each locked in here:
+
+1. create_source() rejects unsupported files at upload time with a 400
+   by running content-core's own identification + routing
+   (ensure_file_type_supported) - deliberately NOT an extension
+   allowlist, so files content-core does handle (e.g. extensionless
+   plain text) keep working.
+2. content_process() re-raises UnsupportedTypeException as ValueError,
+   which is in the command's stop_on list - the job fails on attempt 1
+   instead of retrying a deterministic error.
+3. get_source_status() surfaces the stored error_message for failed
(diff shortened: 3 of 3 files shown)


lfnovo (changes requested):
The approach is right — content-core's own identification+routing instead of an extension allowlist (and the reasoning documented against `tests/test_upload_type_mitigations.py`), the `UnsupportedTypeException` → `ValueError` re-raise so `stop_on` kills the job on attempt 1, and the fail-open posture on the pre-check. The upload-cleanup test locking in the `os.unlink` on rejection is appreciated. Two changes before merge:

1. **Truncate the surfaced error** (`api/routers/sources.py`, `get_source_status`): `error_detail` is the raw `error_message` off the command record, returned to the client unbounded. For the unsupported-type case it's short, but any other failure leaks arbitrary internal exception text. Cap it (`error_detail[:200]`) — this matches how the rest of the API sanitizes error output (cubic flagged this too, and it's correct).

2. **Use the configured document engine in the pre-check** (`ensure_file_type_supported`): it hardcodes `document_engine="auto"` with a comment saying it's the same engine `content_process()` runs with — but the worker actually uses `content_settings.default_content_processing_engine_doc or "auto"` (`open_notebook/graphs/source.py`). With a non-default engine configured, the pre-check can disagree with the worker — including rejecting an upload the worker would process, and fail-open doesn't cover that case (a false `UnsupportedTypeException` is a "successful" rejection, not an unexpected error). Read the same setting the worker reads.

On the content-core internal imports (`content_core.content.extraction.graph`, `content_core.content.identification`): the ImportError guard with fail-open makes this acceptable for now, and I'd rather ship the fix than wait — but the durable home is a public support-check API in content-core. I'm opening an upstream issue for that; no action needed here.

I've also approved the CI workflows — please make sure Tests goes green.

lfnovo:
Closing this as superseded — but the diagnosis here was spot-on and drove the fix that shipped.

Issue #975 is now resolved on `main` (released in v1.13.0) through a different path:

- **#1116** — upgraded the source pipeline to **content-core 2.x**.
- **#1117** — rejects unsupported uploads at ingestion with **HTTP 415**, using content-core 2.x's native public support-check API (`api/routers/sources.py:52-72`).

That's the same fail-fast behavior this PR targeted, implemented against content-core's public support API rather than the internal imports (`content_core.content.extraction.graph` / `.identification`) this branch used — which is exactly the "durable home" I flagged in my earlier review. With the content-core 2.x migration rewriting `source.py`, this branch also now conflicts with `main`.

Thank you @pelazas — your root-cause analysis (the `UnsupportedTypeException` not being in the command's `stop_on` list, so a deterministic failure burned all 15 retries) was correct and directly informed the shipped fix, and the reasoning against an extension allowlist held up. The door's open for future contributions.

The one item from my review that lives on independently of this PR — capping the error text surfaced to clients — I've split out into its own issue.

PR #1046, opened 2026-07-11
fix: actually clear credential fields end to end (frontend payload + API null handling)

lfnovo:
## Summary

Found in v1.11 release testing (bucket C, manual): clearing an Ollama credential's `base_url` in the edit dialog reported success but kept the stale value — the connection test then kept timing out against a dead IP while the field *looked* empty. Investigation revealed **two mirror-image bugs**, one on each side of the wire; either alone makes clearing impossible:

### Frontend (`settings/api-keys/page.tsx`)
The edit submit handler mapped an emptied field to `undefined` (`data.base_url = baseUrl || undefined`). `JSON.stringify` drops `undefined` keys, so the PUT body simply omitted the field. Same pattern on the Vertex `project` / `location` / `credentials_path`. **Fix**: emptied fields are sent as explicit `null`; `UpdateCredentialRequest` typed `string | null` accordingly.

### Backend (`api/routers/credentials.py`)
The update handler guarded every field with `is not None` — so even an explicit `null` was silently skipped and the old value survived a 200 response. Only a hand-crafted `{"base_url": ""}` actually cleared. **Fix**: field updates keyed on presence in `model_fields_set` — absent keeps, explicit `null`/`""` clears. `name`/`modalities`/`api_key` keep their non-null guards (clearing those isn't meaningful).

Pre-existing bugs, not v1.11 regressions — but user-facing, contained, and they ride the release.

## Testing

- Backend regression tests: null-clears, empty-string-clears, absent-keeps, Vertex `credentials_path` — full suite 407 passed
- Frontend: `npm run lint` 0 errors, `tsc --noEmit` clean, vitest 80/80
- Live: after clearing the real credential, the Ollama connection test went from "Connection timed out" to "Connected. 12 models available"

### DIFF
--- a/api/routers/credentials.py
+++ b/api/routers/credentials.py
@@ -216,34 +216,41 @@ async def update_credential(credential_id: str, request: UpdateCredentialRequest
     try:
         cred = await Credential.get(credential_id)
 
+        # Partial-update semantics keyed on field PRESENCE, not value:
+        # a field absent from the payload is left untouched, while an explicit
+        # null (or "") clears it. `is not None` checks would silently ignore
+        # a null sent to clear a field — the old value survived while the
+        # client saw success.
+        sent = request.model_fields_set
+
         if request.name is not None:
             cred.name = request.name
         if request.modalities is not None:
             cred.modalities = request.modalities
         if request.api_key is not None:
             cred.api_key = SecretStr(request.api_key)
-        if request.base_url is not None:
+        if "base_url" in sent:
             cred.base_url = request.base_url or None
-        if request.endpoint is not None:
+        if "endpoint" in sent:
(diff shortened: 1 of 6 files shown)


lfnovo:
Fair point — added in the latest commit: the payload construction was extracted to a pure `buildCredentialUpdatePayload()` and unit-tested. The key test reproduces the original bug precisely: an emptied `base_url` must survive `JSON.parse(JSON.stringify(payload))` with the key present and `null` — with the old `|| undefined` mapping that assertion fails because the key is dropped from the wire format. Also covers the Vertex fields, omit-when-unchanged, and the `num_ctx` clear. 85/85 frontend tests green.

Issue #932, opened 2026-06-20
[Bug]: Note editor `max-h-[500px]` restricts editing space, ignoring `flex-1` parent intent


univerz1977:
### Description

The note editor in the notebook dialog has a hard-coded `max-h-[500px]` that prevents the editor from using the full available vertical space. This contradicts the parent container's `flex-1` layout which is designed to let the editor fill all remaining space in the dialog.

### Location

**File:** `app/frontend/src/app/(dashboard)/notebooks/components/NoteEditorDialog.tsx` (line 165)

### Current Behavior

```tsx
// Parent — intended to fill remaining space:
<div className="flex-1 overflow-y-auto">

  // Child — restricted by hard max-h:
  <MarkdownEditor
    className={cn(
      "w-full h-full min-h-[420px] max-h-[500px] overflow-hidden ...",
      ...
    )}
  />
</div>
```

The `max-h-[500px]` causes two issues:

1. **Wasted space** when the dialog window is resized taller — the editor doesn't grow beyond 500px even though `flex-1` parent expects it to
2. **Limited editing area** for long-form notes — the editor window caps at 500px regardless of content length, requiring excessive scrolling within the editor

### Expected Behavior

The editor should use proportionally more space in the dialog. Since `DialogContent` already has `max-h-[90vh]`, removing the hard pixel cap in favor of a viewport-relative constraint would allow the editor to:

- Use available space when the dialog is large
- Stay bounded when the dialog is small
- Not overflow the page (already handled by `DialogContent`'s `max-h-[90vh]`)

### Suggested Fix

Replace the hard pixel cap with a viewport-relative one:

```diff
- "w-full h-full min-h-[420px] max-h-[500px] overflow-hidden"
+ "w-full h-full min-h-[420px] max-h-[70vh] overflow-hidden"
```

Or alternatively, remove `max-h` entirely and let the parent dialog's `max-h-[90vh]` serve as the sole height constraint.

### Environment

- open-notebook version: v1-latest (`lfnovo/open_notebook:v1-latest`)
- Browser: any
- Dialog: Notebook → "Write Note" button

### How did it break?

open long note.

### Open Notebook Version

v1-latest (Docker)

### Contribution

- [ ] I am a developer and would like to work on fixing this issue (pending maintainer approval)

PR #1047, opened 2026-07-11
fix(frontend): keep the note editor dialog inside the viewport with long content

lfnovo:
## Summary

Found in v1.11 release testing (bucket C, manual): typing a long note made the dialog grow past the viewport with no internal scroll and no reachable Save button.

**This is a regression from #932 in this release** (removing the editor's `max-h-[500px]`): the removal relied on the `flex-1`/`h-full` parent chain, but `DialogContent` isn't a flex container and has no definite height (only `max-h-[90vh]`) — so `h-full` resolved to auto, the `flex-1` wrapper (default `min-height: auto`) never shrank below its content, and the inner `overflow-y-auto` never engaged.

**Fix** — the standard scroll-inside-dialog flex pattern:
- `DialogContent`: `flex flex-col`
- form: `flex-1 min-h-0` (instead of `h-full`)
- editor wrapper: `min-h-0`

The #932 goal is preserved: the editor still fills the dialog on tall windows; it just can't push past it anymore.

## Testing

- Verified in the real browser via Playwright with a 200-line note: dialog height 1089px within a 1210px viewport (≤90vh), footer buttons visible, internal editor scroll active
- `npm run lint` (0 errors), `npx tsc --noEmit` clean, `npm run test` 80/80

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1047?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/app/(dashboard)/notebooks/components/NoteEditorDialog.tsx
+++ b/frontend/src/app/(dashboard)/notebooks/components/NoteEditorDialog.tsx
@@ -120,13 +120,13 @@ export function NoteEditorDialog({ open, onOpenChange, notebookId, note }: NoteE
   return (
     <Dialog open={open} onOpenChange={handleClose}>
       <DialogContent className={cn(
-          "sm:max-w-3xl w-full max-h-[90vh] overflow-hidden p-0",
+          "sm:max-w-3xl w-full max-h-[90vh] overflow-hidden p-0 flex flex-col",
           isEditorFullscreen && "!max-w-screen !max-h-screen border-none w-screen h-screen"
       )}>
         <DialogTitle className="sr-only">
           {isEditing ? t('sources.editNote') : t('sources.createNote')}
         </DialogTitle>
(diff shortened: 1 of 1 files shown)


lfnovo:
Follow-up commit: with only `max-h-[90vh]` the dialog height stayed content-based, so on an **empty** note the percentage height chain collapsed and the typing area shrank to ~90px (reported by Luis while testing). `h-[90vh]` makes the chain definite end to end. Playwright-verified both states: empty note → 895px editor content area; 200-line note → dialog fixed at 1089px, footer visible, internal scroll.

PR #1060, opened 2026-07-11
docs: stop teaching 0.0.0.0 SurrealDB port exposure in setup snippets

### LINKED ISSUE #1034: docs: stop teaching 0.0.0.0 SurrealDB port exposure in setup snippets
#1025 changed the shipped docker-compose.yml to bind SurrealDB to 127.0.0.1 only (the DB starts with root:root, so 0.0.0.0 exposure = anyone reaching the host connects as root). But ~9 docs still embed their own compose / `docker run` snippets that publish `8000:8000` on 0.0.0.0 — and users copy those, not the repo file. One even passes `--bind 0.0.0.0:8000` explicitly.

Known spots to fix (non-exhaustive):
- `docs/0-START-HERE/quick-start-openai.md` (~line 28, includes `--bind 0.0.0.0:8000`)
- `docs/0-START-HERE/quick-start-cloud.md`
- `docs/1-INSTALLATION/docker-compose.md`
- `docs/5-CONFIGURATION/reverse-proxy.md`
- `docs/7-DEVELOPMENT/development-setup.md` (~lines 92, 98, 312)

Fix: bind to 127.0.0.1 in the snippets to match the shipped default, and where remote DB access is genuinely needed, point at the new `docker-compose.override.yml.example` + a firewall/SSH-tunnel note rather than blanket 0.0.0.0 exposure.

Follow-up to #1025.

lfnovo:
## Summary

#1025 changed the shipped `docker-compose.yml` to publish SurrealDB on `127.0.0.1` only, because the database starts with default root credentials — but the docs still embedded their own compose / `docker run` snippets publishing `8000:8000` host-wide, and users copy those. This PR sweeps every setup snippet so nothing teaches exposing the database on `0.0.0.0`:

- **Docs snippets** (README, all four quick starts, installation, reverse-proxy, openai-compatible, advanced, development-setup, dev quick-start): `- "8000:8000"` → `- "127.0.0.1:8000:8000"` and `-p 8000:8000` → `-p 127.0.0.1:8000:8000`, with a short warning comment mirroring the shipped compose file.
- **`--bind 0.0.0.0:8000` dropped** from containerized `surreal start` commands — it is the image's in-container default, and the shipped compose doesn't use it either; the loopback scoping happens on the host-side port publish.
- **`examples/docker-compose-*.yml`** (ollama, full-local, speaches, dev) got the same loopback binding — the installation and local-TTS/STT guides tell users to download these files verbatim.
- **`docker-compose.override.yml.example`** (the sanctioned opt-in re-publish) now uses `ports: !override` — without it, compose merges the override's port list with the base file's and the container fails to start with "port is already allocated" (verified with `docker compose config` against the real files).
- **`docs/5-CONFIGURATION/database.md`** (DB-on-host recipe) gained a note that a loopback-published port isn't reachable at the machine IP, pointing at the override example + firewall/SSH-tunnel guidance.
- The vLLM compose snippet in `openai-compatible.md` now publishes on host port `127.0.0.1:8001` — it previously collided with SurrealDB's host port 8000 in the same compose file (Open Notebook reaches it at `http://vllm:8000/v1` over the compose network regardless).

Closes #1034

## Test plan

- [x] `grep -rn '"8000:8000"|-p 8000:8000|--bind 0.0.0.0'` over `*.md`/`*.yml` — no host-wide publishes remain (except the deliberate, commented override example)
- [x] `docker compose config` on the shipped `docker-compose.yml` + updated override example: merge now yields a single `0.0.0.0:8000` publish (previously two conflicting entries)
- [x] `docker compose -f examples/docker-compose-*.yml config -q` — all four examples still validate
- [x] `uv run python scripts/check_md_links.py` — all relative markdown links resolve
- [x] `uv run pytest tests/` — 407 passed; `ruff check .` clean

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1060?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lfnovo:
Re the remaining P2 (scalar `command` breaking on passwords with spaces): valid, but the docs snippet deliberately mirrors the shipped `docker-compose.yml`, which has the same pattern — fixing only the doc would break that mirror. Split into #1093 to fix both together (YAML list form in the shipped compose, the examples, and the docs snippet).

PR #1065, opened 2026-07-11
refactor(frontend): split api-keys settings page into components

lfnovo:
## Summary

Move-only refactor of `frontend/src/app/(dashboard)/settings/api-keys/page.tsx` (1,441 lines → 147 lines). No user-facing changes.

Each of the six components that lived inside the page file now has its own file under `src/components/settings/` (exported via the existing barrel `index.ts`), and the provider config tables moved to `src/lib/providers.tsx`:

- `CredentialFormDialog.tsx`
- `DiscoverModelsDialog.tsx`
- `DeleteCredentialDialog.tsx`
- `CredentialItem.tsx`
- `ProviderSection.tsx`
- `DefaultModelSelectors.tsx`
- `src/lib/providers.tsx` — `ALL_PROVIDERS`, `PROVIDER_DISPLAY_NAMES`, `PROVIDER_MODALITIES`, `PROVIDER_DOCS`, `TYPE_ICONS`, `TYPE_COLORS`, `TYPE_COLOR_INACTIVE`, `TYPE_LABELS`, and the `ModelType` alias (`.tsx` because `TYPE_ICONS` holds JSX)

Props, names and behavior are unchanged; prop types were only reshaped into `interface XProps` to match the existing convention in `components/settings/` (e.g. `MigrationBanner`).

## Intentional changes (the only two)

1. **Select dedup**: the 47-line default-model `Select` block was duplicated character-for-character between the primary and advanced grids in `DefaultModelSelectors`. It is now a single `DefaultModelSelect` component used by both call sites (advanced passes `showDescription`).
2. **Fossil removal**: dropped the stale comment in `CredentialItem` defending against a Proxy-based `t()` that no longer exists, along with the associated `testModelLabel`/`deleteModelLabel` caching variables — `t()` is called directly now.

## Test follow-up

The backend cross-check test `tests/test_credential_provider_validation.py::TestSupportedProviderMatchesOtherSourcesOfTruth::test_matches_frontend_all_providers_list` parses the frontend source for the `ALL_PROVIDERS` array; it now reads `frontend/src/lib/providers.tsx` instead of the old page file (same regex, new path).

## Verification

Inside `frontend/`:

- `npm run lint` — 0 errors (6 pre-existing warnings, all in untouched files)
- `npm run test` — 12 files, 85 tests passed
- `npm run build` — production build succeeds

Backend: `uv run pytest tests/` — 403 passed.


lfnovo:
Valid catch, but pre-existing: the `.replace('{provider}', ...)` pattern was moved verbatim from the old page file (this PR is move-only by design). There are 74 occurrences of this pattern across the app; an upcoming PR converts them all to i18next interpolation (`t(key, { provider })` + `{{provider}}` in the 14 locales) in one codemod pass. Fixing just this file here would split the pattern migration across two PRs.

PR #1072, opened 2026-07-11
refactor(api): extract shared session and message helpers for chat routers

lfnovo:
## What

`api/routers/chat.py` and `api/routers/source_chat.py` were heavy copy-paste of each other. This extracts the shared pieces into a new `api/routers/_chat_shared.py` and replaces every duplicated site in both routers. **No behavior change** — same status codes, same response shapes, same error messages (including the known missing-relation-becomes-500 quirk, deliberately preserved as-is for a later error-handling PR).

### What was deduplicated

- **`normalize_record_id(table, id)`** — the `x if x.startswith("table:") else f"table:{x}"` idiom, previously inlined **12 times** across both routers (including twice in the same handler: `get_session` and `update_session` recomputed `full_session_id` identically)
- **`get_session_or_404` / `get_source_or_404` / `get_verified_source_session`** — the "verify source exists → normalize session id → fetch session → verify `refers_to` relation" block, copy-pasted **4 times in full and 2 times partially** in `source_chat.py`, plus the session-fetch variant repeated **4 times** in `chat.py`
- **`extract_chat_messages`** — the LangGraph state → `ChatMessage` conversion loop, written **3 times** (2× chat, 1× source_chat; the SSE streaming loop has a different output shape and was left alone)
- **`ChatMessage` and `SuccessResponse`** pydantic models — previously defined twice, now a single definition

### Tests-first approach

Since this is a delicate refactor, characterization tests were written **first** (commit 1) and run green against the unchanged code, then the extraction (commit 2) was verified against them:

- ID normalization edge cases (bare id vs already-prefixed id, for both routers)
- Missing source / missing session → 404 with the exact existing detail strings
- Missing `refers_to` relation → existing behavior pinned as-is: 500 wrapping the inner 404 on get/delete, real 404 on send-message (which re-raises `HTTPException`)
- Message extraction shapes, including the `type`/`content`/positional-id fallbacks and the empty-state path

The only test change in commit 2 is repointing mock patch targets to the new module.

### Line delta

`+156 / −265` in the refactor commit: the two routers shed **250 lines** (chat.py −92, source_chat.py −158) for **94 lines** of shared module.

### Verification

- `uv run pytest tests/` — 415 passed (403 baseline + 12 new)
- `ruff check .` — clean
- `mypy` — 197 errors before and after (the 4 pre-existing errors in these two files are unchanged; none in the new module)

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1072?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lfnovo:
Fixed in 233102e: the three session handlers (get/update/delete) now unpack `_full_source_id`, making explicit that only the verified session is needed there — the source-level verification happens inside `get_verified_source_session()` itself, so there was no enforcement gap. `send_message` keeps using `full_source_id` (it passes it to the graph).

PR #1085, opened 2026-07-11
fix(insights): stamp created/updated on source_insight and stop serializing "None"

### LINKED ISSUE #1045: source_insight records have no created/updated timestamps; API serializes them as the string "None"
Found during v1.11 release smoke testing (pre-existing — not a v1.11 regression).

Two layers:
1. `source_insight` records are saved without `created`/`updated` values — the rows in SurrealDB genuinely hold `NONE` (verified live: `SELECT id, created, updated FROM source_insight`). Other tables get these stamped.
2. The API then wraps them in `str(...)` (`api/routers/insights.py`, `api/routers/sources.py`), so clients receive the literal string `"None"` instead of `null` or an ISO timestamp.

**Impact**: cosmetic/API-hygiene — insight lists can't be sorted by date and clients see `"None"`.

**Suggested fix**: stamp `created`/`updated` on the `source_insight` table (schema default or on save), and make the response models emit `null` when absent instead of `str(None)`.

lfnovo:
Fixes #1045.

## What

`source_insight` was the only content table without `created`/`updated` field definitions. Because the table is SCHEMAFULL and insights are created via a raw `CREATE` (create_insight command), SurrealDB silently dropped any timestamps — rows genuinely held `NONE` — and the API then wrapped them in `str()`, so clients received the literal string `"None"`.

- **Migration 19** defines `created`/`updated` on `source_insight` with the same `time::now()` defaults used by source, note and notebook. Existing rows are left untouched (no backfill).
- `SourceInsightResponse.created/updated` are now `Optional[str]`; routers emit ISO 8601 when present, `null` for legacy rows — never `"None"`.
- Frontend types updated to `string | null`.
- Tests cover the migration definition/registration and the API behavior.

## ⚠️ Why this is a DRAFT — maintainer decision needed

This fix **requires a schema migration (19)**, which broke the no-migration rule of the batch it was developed in. The Python-side alternative was verified impossible: SurrealDB 2.x SCHEMAFULL tables silently drop undefined fields, so stamping from the app layer cannot work. Marking ready-for-review is the explicit approval of shipping migration 19.

Note: branch is based on main as of #1054; needs a rebase before merge (no conflicts expected — the touched files weren't part of the later cleanup PRs, but CHANGELOG will need the usual bullet-merge).

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1085?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lfnovo:
Reviewed the migration and API change — **migration 19 is good to approve.**

Migration correctness:
- **Mirrors the canonical pattern exactly.** `DEFAULT time::now() VALUE $before OR time::now()` (created) and `VALUE time::now()` (updated) are byte-for-byte the definitions `source`/`note`/`notebook` use in migration 1 and `transformation` in migration 5. Same immutable-created / always-bump-updated semantics.
- Registered in `async_migrate.py` on both the up and down lists; `_down` is a correct `REMOVE FIELD IF EXISTS`; `IF NOT EXISTS` keeps it idempotent; number 19 is the next in sequence.
- No backfill is the right call — existing rows return `null` until next touched, and after an update they pick up the same `$before OR time::now()` behavior every other table had at first definition.

API change is type-safe:
- `SourceInsight` inherits `created/updated: Optional[datetime]` (domain `base.py`) with a before-validator coercing DB strings to datetime, so `insight.created.isoformat() if insight.created else None` can't hit the `str`-has-no-`isoformat` trap, and legacy `None` becomes `null` rather than the literal `"None"`.

Optional, non-blocking (cubic P3): the `isoformat() if … else None` branch is duplicated in `insights.py` and `sources.py`. A single Pydantic field serializer (typing the response fields as `Optional[datetime]`) or a small response factory would keep the legacy-null handling in one place. Fine to leave as-is or fold in later.

Under ADR-006 (#1100), the one-migration-per-PR shape here is exactly right. From a review standpoint this is clear to mark ready-for-review — leaving the actual un-draft to you since you flagged it as the maintainer call.

PR #1097, opened 2026-07-12
fix(models): allow clearing optional model defaults

### LINKED ISSUE #1091: Model defaults can't be cleared once set (API ignores null; UI has no "none" option for optional defaults)
## Problem

Once a model default is set, there is no way to clear it:

1. **API**: `PUT /api/models/defaults` ignores `null` values — the update handler only applies fields that are `is not None` (`api/routers/models.py:342`), so a null can never clear a default. Same anti-pattern that #1046 fixed for credential fields (frontend payload + API null handling) — the fix pattern established there applies here.
2. **UI**: the default-model selects (`DefaultModelSelectors`) offer no "none/clear" option.

## Scope note (from maintainer testing)

Not every default should be clearable in the UI: the required ones (chat, embedding — the ones the app can't function without) should arguably never offer "none". Optional defaults (transformation, tools, large-context, TTS, STT) are the ones that need a clear affordance — especially since transformation/tools deliberately fall back to the chat default when unset (`open_notebook/ai/models.py:270-273`), which is behavior a user may want to restore after experimenting.

## Suggested fix

- API: distinguish "field absent" (keep) from "field explicitly null" (clear) — `exclude_unset` semantics, per the #1046 precedent
- UI: add a "Use default fallback (none)" option to the optional defaults only

Found while release-testing v1.12.0 (pre-existing, not a regression).

lfnovo:
## Summary

Once a model default was set, it could never be cleared:

- **API**: `PUT /api/models/defaults` only applied fields that were `is not None`, so an explicit `null` sent to clear a default was silently dropped — the old value survived while the client saw success. This is the same anti-pattern fixed for credential fields in #1046, and the same fix pattern is applied: the handler is now keyed on field **presence** (`model_fields_set`) — a field absent from the payload keeps its current value, an explicit `null` clears it.
- **Required vs optional**: the defaults the app can't function without (`default_chat_model`, `default_embedding_model`) reject `null` with a 400 (`InvalidInputError`) — they can be reassigned, never cleared.
- **UI**: the optional default-model selects (transformation, tools, large context, TTS, STT) now offer a clear option inside the dropdown — labeled **"Use fallback (chat default)"** for transformation/tools (which deliberately fall back to the chat default when unset, `open_notebook/ai/models.py:270-273`) and **"None"** for the rest. The select also *displays* that state when nothing is assigned, instead of the generic "Select a model" placeholder. The transformation default is no longer marked required in the UI, matching its backend fallback behavior. Chat and embedding never offer "None".
- i18n: the two new strings (`models.noneOption`, `models.noneFallbackToChat`) added to all 14 locales.

Closes #1091

## Test plan

- New API tests (`tests/test_models_api.py::TestUpdateDefaultModels`): explicit null clears an optional default; absent field keeps the current value; explicit null on chat/embedding returns 400 and does not persist; required defaults can still be reassigned.
- `uv run pytest tests/`: 466 passed
- `ruff check .` clean · `uv run python -m mypy .` 0 errors
- Frontend: `npm run lint` (0 errors), `npx tsc --noEmit` clean (locale parity via `satisfies TranslationShape`), `npm run test` 92 passed
- Verified callers: the frontend only ever PUTs single-field partials (`{ [key]: value || null }`), and auto-assign mutates the domain object directly (only fills empty slots) — no caller relies on the old "null is ignored" behavior.

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1097?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lfnovo:
Reviewed against current `main` — the implementation is solid. The API switch from `is not None` to `model_fields_set` gives the right presence-based semantics (absent field = keep, explicit null = clear), required slots (chat, embedding) are protected with a clear `InvalidInputError`, and the new i18n keys (`noneOption`, `noneFallbackToChat`) are present across all locales. This matches the #1046 precedent well.

One follow-up worth flagging (not a merge blocker):

**`auto_assign_defaults` will silently re-populate an intentionally-cleared optional slot.** `auto_assign_defaults` (`api/routers/models.py:735`) treats any empty slot as "missing" and assigns the first available model. So once this ships, a user who deliberately clears e.g. `default_tools_model` to fall back to the chat default will have it re-filled the next time Auto-assign runs. This isn't a regression (auto-assign already behaved this way), but it does undercut the "clear to restore fallback" behavior this PR introduces. Consider either skipping the fallback-capable optional slots in auto-assign, or distinguishing "intentionally cleared" from "never set". Fine to handle as a separate issue.

Issue #1031, opened 2026-07-10
refactor: migrate database layer to lfnovo/surreal-basics


lfnovo:
Planned migration of the SurrealDB connection + migration layer from the in-repo implementation (`open_notebook/database/repository.py` + `async_migrate.py`) to [lfnovo/surreal-basics](https://github.com/lfnovo/surreal-basics), which brings proper connection management (today every `repo_query` opens a fresh connection + signin), SurrealDB v3 support, and its own migration runner/CLI.

## Migration-compat notes gathered so far

- **Tracking table name already matches**: both use `_sbl_migrations`.
- **Record shape does NOT match**: open-notebook writes `type::thing('_sbl_migrations', $version)` with fields `version` + `applied_at` only; surreal-basics' `get_applied_versions()` reads `r["name"]` with no fallback and defines the table SCHEMAFULL with a required `name` field. Existing installs would KeyError on the first status read — the cutover needs a small data migration backfilling `name` (and note `DEFINE TABLE IF NOT EXISTS` will skip on the pre-existing schemaless table).
- **Record ID shape differs**: ours are deterministic (`_sbl_migrations:<version>`), the lib's are random with a UNIQUE index on `version`. Harmless for reads, relevant if the lib ever upserts by id.
- **Concurrent-replica version-recording crash exists in both implementations**: fixed upstream via lfnovo/surreal-basics#24 rather than patching the code being replaced here (see PR #1022, closed in favor of this plan).
- Open-notebook's `repo_query` also does `parse_record_ids()` normalization (RecordID → string) that several call sites rely on (e.g. batch key matching in insights/podcasts) — verify surreal-basics' return shape or keep a thin adapter.

## Suggested steps
1. Land lfnovo/surreal-basics#24 upstream
2. Adapter layer: swap `repo_query`/`repo_create`/etc. internals to surreal-basics connections, keeping the current function signatures
3. Tracking-table backfill migration (`name` field)
4. Swap AsyncMigrationManager for the lib's AsyncMigrationRunner
5. Remove the in-repo implementation

lfnovo:
**Query-audit item for the v3 upgrade** (from PR #1023): `api/routers/embedding_rebuild.py`'s mode="existing" estimate uses `SELECT VALUE count(...) as count FROM {}` — an empty-object FROM that works on SurrealDB v2 only by parser tolerance (verified live on v2.6.0: it does return the correct count today, so it's not a live bug). The documented-semantics replacement, also verified live, is:

[...]

PR #1100, opened 2026-07-12
docs: record ADR-006 — migration granularity follows merge granularity

lfnovo:
Documents the migration policy discussed today, prompted by PR #1085's migration-19 question:

- **One migration per PR that needs one**, reviewed with the code that motivates it, with its `_down` counterpart
- **Numbers allocated in merge order** — strict mode surfaces collisions between parallel PRs as a required update-branch
- **Never consolidate after a migration lands on main**: `v1-dev` images publish on every push, so a migration is effectively released when merged — post-hoc squashing would desync `_sbl_migrations` for dev-image users
- Consolidation is a **development-time** choice only, when in-flight PRs touch the same table

Also cross-references the ADR from the Database Migration playbook and adds the missing ADR-005 row to the decisions index.

## Testing
- `scripts/check_md_links.py` passes (all relative links resolve)

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1100?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/7-DEVELOPMENT/change-playbooks.md
+++ b/docs/7-DEVELOPMENT/change-playbooks.md
@@ -122,6 +122,7 @@ Step-by-step guides for common types of changes in the Open Notebook codebase. E
 **Important:**
 - Migrations are numbered and run in order
 - They're tracked in the `_sbl_migrations` table — won't re-run
+- One migration per PR that needs one, numbered in merge order; never consolidate after a migration lands on main (dev images apply it immediately) — see [ADR-006](decisions/ADR-006-migration-granularity.md)
 - For destructive changes (DROP FIELD), consider data preservation
 - Test with existing data, not just empty database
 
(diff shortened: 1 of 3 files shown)


lfnovo:
Reviewed — approve. The ADR is sound and accurately reflects how this project actually ships: migrations run at startup in sequence via `_sbl_migrations`, and `v1-dev` publishes on every push, so "released on merge to main" is the correct mental model and the "never consolidate after main" rule follows directly from it. The alternatives section captures the real failure mode (post-hoc squashing desyncs `_sbl_migrations` for dev-image users).

Checks:
- ADR-005 file exists, so the new index row in `decisions/README.md` resolves (good catch backfilling that missing row).
- The `change-playbooks.md` cross-reference points at the right ADR path.
- Policy correctly noted as runner-agnostic, so it survives the surreal-basics move (#1031).

No changes requested. This is the written form of the migration-19 decision, so it pairs naturally with #1085.

PR #1102, opened 2026-07-12
fix(frontend): shared "content no longer exists" state for orphan references

### LINKED ISSUE #455: [Bug]: Handle orphan references better

When we delete a document (source, insight or note), the chat will fail to handle the orphan reference: 

Image

### How did it break?

Find a more elegant solution to handle this. 

### Logs or Screenshots

_No response_

### Open Notebook Version

v1-latest (Docker)

### Environment

_No response_

### Additional Context

_No response_

### Contribution

lfnovo:
## Summary

When a chat/Ask citation references a source, insight, or note that has since been deleted, clicking it now opens a single, friendly "this content no longer exists" state in all three dialogs — instead of today's inconsistent behavior (generic red error for sources, silently blank dialog for insights, empty ghost editor inviting edits for notes).

Frontend-only, per the settled design on the issue: backend 404s were already clean for all three GET-by-id endpoints, and dangling refs only exist in persisted message text, discovered at click time.

### Changes

- **Shared state**: new `ContentUnavailable` component (built on the existing `EmptyState`) with two variants — `not-found` (HTTP 404 → "deleted / no longer exists") and `error` (transient failure → "unable to load, try again") — used identically by `SourceDetailContent`, `SourceInsightDialog`, and `NoteEditorDialog`.
- **Insight dialog**: consumes `isError` from `useInsight`; on error all derived fields (content, badge, "View source") go blank through a single gate point instead of falling back to the blank stub.
- **Note editor**: consumes `isError` from `useNote`; the not-found state fully gates the editor and its save path, killing the ghost-edit flow.
- **Source dialog**: distinguishes 404 from other errors (previously both showed a red "failed to load"); `SourceDetailContent` is keyed by `sourceId` so loading/not-found state resets per source. The file-download handler now reuses the new `isNotFoundError` helper.
- **No retry on 404**: the global react-query retry predicate skips 404s (`retry: (failureCount, error) => !isNotFoundError(error) && failureCount < 2`) — retrying cannot resurrect a deleted item; non-404 behavior is unchanged.
- **i18n**: new `common.contentUnavailable.*` strings added to all 14 locales; now-unused `sources.notFound` and `sources.loadFailed` keys removed everywhere (enforced by the locale parity/unused-key tests and `satisfies TranslationShape`).

Out of scope by design: pre-marking orphan refs in message text (would cost N validation requests per rendered message) and the delete-while-editing race (already surfaces a save error).

Closes #455

## Test plan

- New component tests (mocking hooks/API to return axios 404 and network errors):
  - `SourceInsightDialog.test.tsx` — not-found state, error state, close affordance, success render
  - `NoteEditorDialog.test.tsx` — not-found/error states gate the editor and save button, close affordance, success render
  - `SourceDetailContent.test.tsx` — 404 vs non-404 variants, onClose wiring
- `npm run test` — 103 passed (includes locale parity + unused-key tests validating the i18n changes)
- `npx tsc --noEmit`, `npm run lint`, `npm run build` — clean
- `uv run pytest tests/` — 475 passed (backend untouched, hygiene run)

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1102?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
+- Clicking a chat/Ask citation that points at a deleted source, insight, or note now shows a shared, friendly "this content no longer exists" state in all three dialogs (instead of a raw error, a blank dialog, or an empty editable note editor), 404 lookups are no longer retried, and non-404 failures show a distinct "unable to load" message (#455)
 - Source insights now get `created`/`updated` timestamps stamped at creation (migration 19 mirrors the defaults used by the other tables), and the insights API returns `null` — instead of the literal string `"None"` — for legacy insights that predate the migration (#1045)
 - `uv sync` alone now provides the full dev toolchain: the legacy `[project.optional-dependencies].dev` list was merged into `[dependency-groups].dev` (mypy included — the documented `uv run python -m mypy .` previously failed on a fresh clone), Jupyter-only packages moved to a separate `notebooks` group, and the CI typecheck job no longer needs `--extra dev` (#1101)
 - Optional model defaults (transformation, tools, large context, TTS, STT) can now be cleared: `PUT /api/models/defaults` honors explicit `null` (field absent still means "keep"; chat and embedding defaults reject `null`), and the default-model selects offer a "None" / "Use fallback (chat default)" option for the optional defaults (#1091)
--- a/frontend/src/app/(dashboard)/notebooks/components/NoteEditorDialog.test.tsx
+++ b/frontend/src/app/(dashboard)/notebooks/components/NoteEditorDialog.test.tsx
@@ -1,0 +1,117 @@
+import { render, screen, within } from '@testing-library/react'
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
+import { describe, it, expect, vi, beforeEach } from 'vitest'
+import { NoteEditorDialog } from './NoteEditorDialog'
+import { useNote } from '@/lib/hooks/use-notes'
+
+// useTranslation is mocked globally in setup.ts (t returns the key string)
+
+vi.mock('@/lib/hooks/use-notes', () => ({
+  useNote: vi.fn(),
+  useCreateNote: () => ({ isPending: false, mutateAsync: vi.fn() }),
+  useUpdateNote: () => ({ isPending: false, mutateAsync: vi.fn() }),
+}))
+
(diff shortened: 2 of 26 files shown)


lfnovo:
Reviewed — approve. Clean, consistent, and well-tested. The shared `ContentUnavailable` (404 → "no longer exists" vs transient → "unable to load") is wired identically into all three dialogs, each gating through a single point:

- **Insight**: `displayInsight = isError ? undefined : (fetchedInsight ?? insight)` blanks every derived field (including "View source") on error — no more silent blank dialog.
- **Note**: `noteUnavailable = isEditing && noteError` gates the whole form and its save path, killing the ghost-edit flow; create flow is untouched.
- **Source**: `loadError` is now typed `'not-found' | 'error'`, 404 is distinguished from transient failures, and `key={sourceId}` on both mount points resets per-source state on navigation. The download handler reuses `isNotFoundError` too.

Verified:
- **i18n complete** — the 4 new `common.contentUnavailable.*` keys plus `common.close` exist in all 14 locales.
- **Tests are thorough** — each of the three components covers 404 → not-found, non-404 → error, close-button behavior, and successful load.
- The `query-client` `retry` change (never retry a 404) is a sensible global policy; just flagging that it applies to *all* queries, not only these three — intended and fine.

One follow-up (cubic P2, non-blocking, **out of scope here**): `SourceDetailContent` uses imperative `fetchSource` + local `loading`/`loadError` state, whereas the insight/note dialogs use the `useInsight`/`useNote` React Query hooks — and a `useSource(id)` hook already exists. Migrating the source path to `useSource` would unify retry/cache/error handling and remove the need for the parent `key` remount. But that imperative pattern predates this PR (this change only adapted its error handling), so the `key` remount is the right minimal fix for this bug; the migration is better as a separate cleanup that makes all three dialogs consistent.

Nit (optional): in `SourceDialog`, `SourceDetailContent` doesn't appear to receive `onClose`, so the not-found state there renders without its own close button — the dialog's own dismiss still works, so harmless, just slightly less consistent than the insight/note dialogs.

lfnovo:
Filed the `useSource` migration follow-up as #1106 — tracked separately so it doesn't hold up this fix. We'll pick it up later, next time this area is touched.

PR #1109, opened 2026-07-12
refactor(frontend): migrate SourceDetailContent to the useSource hook

### LINKED ISSUE #1106: Migrate SourceDetailContent to the useSource hook for consistency with insight/note dialogs
## Context

Follow-up from #1102 (shared "content no longer exists" state for orphan references).

That PR made all three reference dialogs handle deleted/errored content consistently at the UI level, but they diverge underneath:

- **Insight dialog** → `useInsight` (React Query hook)
- **Note dialog** → `useNote` (React Query hook)
- **Source** (`SourceDetailContent`) → imperative `fetchSource` with local `loading` / `loadError` state

The imperative fetch in `SourceDetailContent` predates #1102, so that PR worked within it and added a `key={sourceId}` remount on the parent (`SourceDialog`, `sources/[id]/page.tsx`) to reset per-source loading/not-found state. That's a fine minimal fix, but it introduces a parent-reset contract that a query-hook-based component wouldn't need.

## Proposal

Migrate `SourceDetailContent`'s source fetch to the existing `useSource(id)` hook (`frontend/src/lib/hooks/use-sources.ts:79`), classifying its error for `ContentUnavailable` the same way the insight/note dialogs already do (`isNotFoundError(error) ? 'not-found' : 'error'`).

Benefits:
- Retry/cache/error behavior lives in one place (React Query) instead of hand-rolled local state — and picks up the global "never retry 404" policy from #1102 automatically.
- Removes the need for the `key={sourceId}` remount contract, since the query re-runs on id change.
- Makes all three reference dialogs consistent.

## Scope note

Cosmetic/maintainability, not user-facing — #1102 already fixed the actual UX. No rush; do it when touching this area next.

## Acceptance criteria

- `SourceDetailContent` derives `source` / loading / error from `useSource(sourceId)`; the local `loading` and `loadError` state and the imperative `fetchSource` are gone.
- 404 → `ContentUnavailable variant="not-found"`, other errors → `variant="error"`, matching current behavior.
- The `key={sourceId}` remounts added in #1102 can be removed (state resets via the query) — verify per-source navigation still resets correctly.
- Existing `SourceDetailContent` tests still pass (adjusted for the hook if needed).

lfnovo:
## Summary

Migrates `SourceDetailContent` from an imperative `fetchSource` with local `loading`/`loadError` state to the shared `useSource(sourceId)` React Query hook, making all three reference dialogs (insight, note, source) consistent underneath, per the acceptance criteria in the issue.

- `source`/loading/error now derive from `useSource(sourceId)`; the local `loading`/`loadError` state and `fetchSource` are gone. Error classification matches the insight/note dialogs: `isNotFoundError(error) ? 'not-found' : 'error'` → `ContentUnavailable`. The never-retry-404 policy comes from the global query-client predicate.
- The `key={sourceId}` remounts added in #1102 on `SourceDialog` and `sources/[id]/page.tsx` are removed. Instead of a fragile enumerated reset effect, the component keys an inner subcomponent by `sourceId` itself: **all** per-source UI state (active tab, in-flight flags like `isDownloadingFile`/`isEmbedding`, insight selection, `fileAvailable`) still resets on per-source navigation, without the parent-reset contract and without a one-frame stale flash on cached navigations.
- Cache correctness now that the detail view is cache-backed:
  - Deleting from the detail view goes through `useDeleteSource` (invalidates `['sources']` incl. the detail key) — previously a raw `sourcesApi.delete` would have let a reopened dialog serve the deleted source from the cache within `staleTime`.
  - Title edits go through `useUpdateSource` (source lists refresh too, which the old local-state update never did), plus a `setQueryData` patch so the title updates instantly while the invalidation refetch is in flight.
  - `ContentUnavailable` only replaces the view when there is no data, so a failed background refetch (`refetchOnWindowFocus`) doesn't blank out a rendered source.
  - `file_available` is synced from the query payload via an effect (downloads can still flip it locally on 404).

No backend changes. No new i18n strings (reuses #1102's).

Closes #1106

## Test plan

- `npm run test` — 103 tests / 17 files pass, including the existing `SourceDetailContent` not-found / error / onClose tests (already rendered under a `QueryClientProvider` mocking `sourcesApi.get`, so they exercise the hook path unchanged)
- `npm run lint` — 0 errors (6 pre-existing warnings in untouched files)
- `npx tsc --noEmit` — clean
- `npm run build` — compiles
- `uv run pytest tests/` — 475 passed (backend hygiene, untouched)

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1109?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lfnovo:
Reviewed — one real bug to fix before merge, otherwise the migration is clean and does exactly what #1106 asked (derives from `useSource`, drops the `key` remounts, routes delete/update through `useDeleteSource`/`useUpdateSource` for cache correctness, keys an inner subcomponent by `sourceId` to reset per-source UI state).

## 🔴 Blocker — deleted sources can render stale (regression of the #1102 scenario)

The render gate changed from `if (loadError || !source)` to `if (!source)` (line 401). React Query **retains the last `data` when a refetch fails**, so this reintroduces exactly the orphan-reference hole #1102 closed:

- `useSource` here is `staleTime: 30s` + `refetchOnWindowFocus: true`, and `useDeleteSource` does `invalidateQueries(['sources'])` (invalidate, not remove).
- Reachable path: view source A → A gets deleted → reopen A from a dangling chat/ask citation within the cache window (or just refocus the tab). The query serves cached data (`source` truthy), background-refetches, gets a 404 (not retried, per the global predicate), and **keeps the stale data** while setting `error`. So `source` is truthy and `loadError === 'not-found'`.
- `isPending` is false (we have data), and `if (!source)` is false → the component falls through and renders the **deleted** source instead of the shared "content no longer exists" state.

Suggested fix — gate on not-found even when stale data lingers, while still tolerating a transient refetch error over good data:

```tsx
if (loadError === 'not-found' || !source) {
  return <ContentUnavailable variant={loadError ?? 'error'} onClose={onClose} />
}
```

That's actually a touch better than the pre-migration `loadError || !source`, which would also blow away good cached content on a transient background failure. A test for "cached source, refetch returns 404 → shows not-found" would lock this in.

## Otherwise 👍

- Deriving `source`/`isPending`/`error` from `useSource` and classifying the error the same way as the insight/note dialogs is the right consistency win.
- Removing the `key={sourceId}` remounts in favor of keying an inner subcomponent reads well and avoids the one-frame stale flash.
- Routing deletion through `useDeleteSource` (cache invalidation) instead of a raw `sourcesApi.delete` is the correct cache-correctness fix now that the view is cache-backed.

Once the gate handles the truthy-stale + not-found case, this is good to go.

lfnovo:
Fixed in 676c4c7 — thanks for catching the truthy-stale + not-found hole.

- The render gate is now `if (loadError === 'not-found' || !source)` exactly as suggested: a definitive 404 always yields `ContentUnavailable` even when React Query retains stale data from a failed refetch, while a transient background error over good cached data still keeps the source rendered.
- Added the regression test: a stale cache entry is seeded via `queryClient.setQueryData(..., { updatedAt: now - 60s })` so mounting triggers a refetch that rejects with 404 (React Query keeps the previous data alongside the error), and the test asserts the shared not-found state renders instead of the cached source.

Gates re-run: `npm run lint` (0 errors), `npm run test` (104 passed), `npx tsc --noEmit`, `npm run build` — all green.

lfnovo:
Re-reviewed 676c4c7 — the fix is exactly right, approve. The gate is now `if (loadError === 'not-found' || !source)` with a clear comment on why a definitive 404 wins over stale cached data while a transient error keeps the good render. And the regression test nails the scenario: it seeds `QUERY_KEYS.source(...)` with a cached entry stamped 60s old (past the 30s `staleTime`) so mount triggers a refetch that rejects 404, then asserts the not-found state shows and the stale title is gone. That's precisely the orphan-reference path. Good to merge.

Issue #1044, opened 2026-07-11
Podcast generation ignores the speaker_profile parameter


lfnovo:
Found during v1.11 release smoke testing (pre-existing, dates back to #93/#632 — not a v1.11 regression).

`POST /api/podcasts/generate` accepts a `speaker_profile` parameter, but `generate_podcast_command` (`commands/podcast_commands.py`) ignores it and always re-derives the speaker via `SpeakerProfile.get_by_name(episode_profile.speaker_config)`.

**Impact**: the UI is unaffected (`GeneratePodcastDialog` always sends `speaker_config` derived from the selected episode profile), but direct API callers passing a valid `speaker_profile` get it silently ignored. If the episode profile's own `speaker_config` points at a renamed/deleted speaker profile, generation fails with "Speaker profile 'X' not found" even when the caller supplied a valid one.

**Repro**: create episode profile A with `speaker_config: "old-name"`, rename the speaker profile to "new-name", then `POST /api/podcasts/generate` with `episode_profile: "A", speaker_profile: "new-name"` → job fails.

**Suggested fix**: honor `input_data.speaker_profile` when provided, falling back to `episode_profile.speaker_config` — or drop the parameter from the API contract.

PR #1110, opened 2026-07-12
fix(podcasts): reference speaker profiles by record ID instead of name

### LINKED ISSUE #630: Episode profile references speaker profile by name instead of ID
## Problem

The `episode_profile.speaker_config` field stores the speaker profile **name** as a plain string instead of referencing it by record ID (`record<speaker_profile>`). This means renaming a speaker profile breaks all episode profiles that reference it.

## Current behavior

- DB schema: `DEFINE FIELD speaker_config ON TABLE episode_profile TYPE string`
- Pydantic model: `speaker_config: str`
- Resolution is done by name lookup: `SpeakerProfile.get_by_name(episode_profile.speaker_config)`

## Expected behavior

`speaker_config` should be a `record<speaker_profile>` reference, similar to how `outline_llm` and `transcript_llm` reference `record<model>`. This would:
- Prevent broken references when renaming speaker profiles
- Be consistent with the model registry pattern
- Allow SurrealDB to enforce referential integrity

## Scope

- DB migration to change field type from `string` to `record<speaker_profile>`
- Data migration to convert existing name strings to record IDs
- Backend model update (`EpisodeProfile.speaker_config`)
- API schema update
- Frontend form update (speaker selector should use ID instead of name)
- `podcast_commands.py` update (load by ID instead of name)

lfnovo:
## Summary

Renaming a speaker profile silently broke every episode profile that referenced it, because `episode_profile.speaker_config` stored the speaker profile **name**. This PR converts the field to a `record<speaker_profile>` reference, per the design settled in the issue discussion.

Closes #630

### What changed

- **Migration 20** (`20.surrealql` + `20_down.surrealql`, registered in `AsyncMigrationManager`): widens the field type, converts existing rows by looking up the speaker profile with the matching name and storing its record ID, then tightens the type to `option<record<speaker_profile>>`.
- **Backend**: `EpisodeProfile.speaker_config` is now an optional record-ID string (nullable, `ensure_record_id` on save, matching the existing `voice_model` pattern). New `SpeakerProfile.resolve()` accepts a record ID or a name. `PodcastService.submit_generation_job` resolves the user-facing name to a record ID at the API boundary and submits the command with the ID; `generate_podcast_command` resolves either form and keeps the #1044/#1058 precedence (explicit `speaker_profile` param wins, `speaker_config` is the fallback).
- **Episode profiles API**: responses return both `speaker_config` (record ID, nullable) and `speaker_config_name` (resolved name for display). Create/update accept a record ID and still accept a name for backward compatibility (resolved server-side; unknown references return 400).
- **podcast-creator integration**: the profile dicts handed to `configure()` get `speaker_config` rewritten back to the speaker **name** (podcast-creator's contract), and profiles whose reference no longer resolves are dropped from the config with a warning — so a single orphaned profile cannot fail validation for every generation.
- **Frontend**: the episode-profile form selects speakers by record ID; all displays resolve the ID to a name (`speaker_config_name`); `GeneratePodcastDialog` sends the resolved name to `POST /api/podcasts/generate` and blocks with a clear toast when the reference is unresolvable. New i18n keys added to all 14 locales.

### API contract

`POST /api/podcasts/generate` is **unchanged**: `speaker_profile` stays a name string (the semantics #1058 cemented). The name→ID resolution happens at the boundary.

### Migration notes (orphaned references)

- An episode profile whose stored name no longer matches any speaker profile at migration time gets `speaker_config = NONE`. **Decision: null it rather than keep the stale string** — the app treats a missing speaker as "needs setup" (UI shows it and asks the user to pick a speaker; generation fails with a clear, actionable error). Keeping the dead name would only preserve a value nothing can resolve, and the record type forbids strings anyway.
- SurrealDB record links do **not** prevent dangling references (deleting a speaker profile later leaves the link in place), so all resolution paths handle "reference doesn't resolve" the same as null; #1044's explicit-parameter fallback remains as defense in depth.
- The down migration converts record IDs back to names and settles on `option<string>` (not the original required `string`), because rows orphaned by the up migration have no name to restore.
- Migration SQL (up and down, exactly as the runner executes it) was validated against a throwaway local SurrealDB 2.6 instance: names convert to IDs, orphans become null, string writes are rejected after tightening, and the rollback restores names.

## Test plan

- `uv run pytest tests/` — 484 passed. New/updated coverage in `tests/test_podcast_speaker_profile.py`:
  - explicit `speaker_profile` param still wins over `speaker_config` (#1044/#1058 regression, updated for record IDs)
  - fallback to `speaker_config` record ID when the param is omitted
  - orphaned (`None`) reference fails with a clear "no speaker profile configured" error before any lookup
  - `SpeakerProfile.resolve` dispatch (record ID → direct lookup, otherwise name lookup, `None` when missing)
  - API boundary submits the resolved record ID to the command
  - one orphaned profile does not poison podcast-creator's episode config, and the config carries speaker names again
  - migration 20 content + registration in `AsyncMigrationManager` (hard-coded list)
- `ruff check .` clean, `uv run python -m mypy .` clean (120 files)
- Frontend: `npm run lint` (no errors), `npm run test` (92 passed), `npx tsc --noEmit` clean, `npm run build` succeeds

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1110?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lfnovo:
Reviewed — solid, and the migration is genuinely well-built. Two things worth addressing (one mechanical, one a should-fix), neither large.

## Migration 20 ✅

- **Convert-in-place is correct and the orphan path is safe.** Widen → `UPDATE ... SET speaker_config = (SELECT VALUE id FROM ONLY speaker_profile WHERE name = $parent.speaker_config LIMIT 1) WHERE type::is::string(...)` → tighten to `option<record<speaker_profile>>`. I ran the risky idiom against the dev SurrealDB: `SELECT VALUE id FROM ONLY ... WHERE name = '<no match>' LIMIT 1` returns `NONE`, not an error — so an orphaned name becomes `NONE` exactly as documented, no migration failure.
- Down migration is right too: widens, `speaker_config.name` back-fill (dangling links → NONE), settles on `option<string>` (correctly relaxed from the original required string, since orphans have no name to restore).
- Registered up+down in `AsyncMigrationManager`; number 20 is correct — 19 (#1085) is already on main.
- The comment even flags that SurrealDB record links don't prevent dangling references, and the app resolves defensively — matches `SpeakerProfile.resolve()` returning None for unresolvable refs.

## Backend / API ✅

- `resolve()` accepting id-or-name, `ensure_record_id` on save, the `_prepare_save_data` pattern — all consistent with `voice_model`/`outline_llm`.
- Command precedence preserved (`input_data.speaker_profile or episode_profile.speaker_config`, #1044/#1058), unknown refs → `InvalidInputError` (400), and the podcast-creator adapter drops unresolvable profiles so one orphan can't fail the whole config.
- i18n complete — `speakerProfileMissing`/`speakerProfileMissingDesc` in all 14 locales.

## Should-fix (cubic P2, verified) — generation still submits the speaker by name

`GeneratePodcastDialog` submits `speaker_profile: selectedEpisodeProfile.speaker_config_name` (the display name) even though it already holds the stable `speaker_config` record ID. Since `SpeakerProfile.resolve()` accepts IDs, submitting `speaker_config` (the ID) would be strictly more robust. As written, renaming a speaker profile while the podcast dialog is open makes generation fail on the stale cached name — a narrower version of exactly the rename-fragility this PR sets out to remove (#630). One-liner: submit the ID, keep `speaker_config_name` for display only.

## Optional (cubic P2) — command mixes orchestration with schema translation

`generate_podcast_command` now inlines the DB→podcast-creator adapter (record-ID→name rewrite, orphan filtering, selected-profile special case). The logic is correct; it's just a lot living in the command path. Extracting a focused materialization helper would keep future profile-reference changes from expanding this function. Non-blocking.

## Mechanical

Branch is `CONFLICTING` — CHANGELOG.md only (union file). Needs a rebase before merge; trivial.

Fix the dialog to submit the ID and rebase, and this is good to go. The migration itself I'm confident in.

lfnovo:
Addressed the review:

- **Should-fix — dialog now submits the record ID**: `GeneratePodcastDialog` sends `speaker_profile: selectedEpisodeProfile.speaker_config` (the stable record ID) instead of the cached display name, so renaming a speaker profile while the dialog is open can no longer break generation. `speaker_config_name` is display-only now. The pre-submit guard blocks both the orphaned case (`speaker_config` null) and the dangling case (`speaker_config_name` null because the reference stopped resolving) with the existing toast. No test asserted the submitted payload shape, so none needed adjusting; the `PodcastGenerationRequest` type doc was updated to say the field carries a record ID (name still accepted).
- **Rebase**: merged `origin/main` into the branch; the only conflict was `CHANGELOG.md` and both sides' entries were kept.
- The optional materialization-helper extraction was left out as agreed (non-blocking).

Gates re-run after the merge: `uv run pytest tests/` (483 passed), `ruff check .`, `mypy` (clean), and in `frontend/`: `npm run lint`, `npm run test` (116 passed), `npx tsc --noEmit`, `npm run build` — all green.

lfnovo:
Re-reviewed the follow-ups (e565d14) — approve.

- **Should-fix resolved**: the dialog now submits `speaker_profile: selectedEpisodeProfile.speaker_config` (the stable record ID), so a rename while the dialog is open can't break generation — exactly the point of #630. The backend resolver takes the ID directly.
- **Nice touch on the guard**: `if (!speakerProfileRef || !selectedEpisodeProfile.speaker_config_name)` blocks both failure modes before submit — orphaned-by-migration (no ID) and dangling reference (ID present but the speaker was deleted, so no resolved name) — with the `speakerProfileMissing` toast.
- CHANGELOG conflict cleared via the origin/main merge; `MERGEABLE` now.

The optional adapter-extraction (P2) is fine to leave for later. Migration 20 I already validated (orphan path returns NONE against the live DB), numbering is correct now that 19 is on main. Good to merge.

PR #1111, opened 2026-07-12
refactor(podcasts): store audio paths relative to PODCASTS_FOLDER

### LINKED ISSUE #1030: refactor: store podcast audio paths relative to PODCASTS_FOLDER
Follow-up to #1018, which added read-time containment guards (`_is_audio_path_contained()`) at every consumption point of `PodcastEpisode.audio_file` (stream, list, delete, retry).

Those guards are the right immediate hardening, but the root cause is that `audio_file` stores an **absolute** path (`/data/podcasts/episodes/<uuid>/audio.mp3`). That has two structural defects:

1. **Path traversal stays representable.** A tampered/buggy record can point anywhere on disk, and safety depends on every current and future endpoint remembering to apply the guard. Storing the path **relative** to `PODCASTS_FOLDER` (`episodes/<uuid>/audio.mp3`) and joining + resolving + containing at a single choke point makes the whole bug class unrepresentable.
2. **Legacy breakage on data-folder moves.** Absolute paths fossilize the deployment layout into the database: relocating `DATA_FOLDER` breaks playback of every previously generated episode. Relative paths survive relocation for free.

Scope sketch:
- Write `audio_file` as a relative path in the podcast generation command (plus validate on write, so the DB never even holds a bad value)
- Migration converting existing absolute rows to relative (strip the `PODCASTS_FOLDER` prefix; rows that don't match the prefix are exactly the out-of-root legacy cases #1018's guards now 403 — decide whether to keep or flag them)
- Resolve + contain in one shared domain/service helper instead of per-router guards
- Release note for the behavior change

Not urgent — #1018's guards cover the risk today. This is the "make it impossible" version.

lfnovo:
## Summary

Follow-up to #1018, which added read-time containment guards at every consumption point of `PodcastEpisode.audio_file`. This PR fixes the root cause: `audio_file` stored an **absolute** path (sometimes as a `file://` URI), which (a) kept path traversal representable in the database and (b) fossilized the deployment layout — moving `DATA_FOLDER` broke playback of every previously generated episode.

Closes #1030

### What changed

- **Relative storage + write-time validation** — the generation command (`commands/podcast_commands.py`) now stores `audio_file` relative to `PODCASTS_FOLDER` (`episodes/<uuid>/audio/<name>.mp3`). `to_relative_audio_path()` validates on write and raises `ValueError` (permanent job failure) if the generated file is outside the podcasts root, so the DB can never hold an absolute or escaping value. `build_episode_output_dir()` now builds from `PODCASTS_FOLDER` directly, so the write root and the validation root cannot drift.
- **Single choke point** — `open_notebook/podcasts/audio_paths.py` replaces the per-router `_resolve_audio_path`/`_is_audio_path_contained` guards. `resolve_contained_audio_path()` joins the stored value with `PODCASTS_FOLDER`, resolves symlinks/`..`, and verifies containment. All five consumption points (stream, list, get, delete, retry) go through it. Absolute paths and `file://` URIs are treated as legacy-invalid — same 403 / `audio_url: null` behavior #1018's guards give out-of-root rows today.
- **In-band error handling** (surfaced by review) — podcast-creator reports audio-combination failures as an `"ERROR: ..."` string in `final_output_file_path`. Previously that garbage string was stored as the audio path with the job reporting success; now the transcript/outline are still persisted and the job fails with the real ffmpeg/clip error.

### Migration notes (migration 21)

Static SurrealQL in the numbered chain (per ADR-006 and the Database Migration playbook). A SurrealQL migration cannot read env vars, but `DATA_FOLDER` is hardcoded to `./data`, so the real root only varies with the process CWD — and the shipped Docker layout (`WORKDIR /app`) is a known constant. The migration converts by known-prefix stripping, in two statements:

1. Normalize plain `file:///` URIs (no percent-encoding, empty host) to absolute paths — SurrealQL cannot URL-decode, so percent-encoded or `file://host/...` forms stay untouched (legacy-invalid).
2. Strip **exactly one** root prefix per row through a single `IF/ELSE` chain (a stripped remainder can never be stripped twice): `/app/data/podcasts/` (shipped Docker layout), `/data/podcasts/` (older Docker layout), `./data/podcasts/` and `data/podcasts/` (CWD-relative writer forms).

Per the settled design decision on the issue: rows under any **other** root (e.g. a source checkout at a custom absolute path) stay untouched — the resolve helper treats any non-relative value as legacy-invalid, preserving today's 403/404 behavior. No nulling, no flags. Trade-off worth being explicit about: non-Docker checkouts whose old rows hold machine-specific absolute paths lose playback of those episodes until regenerated (called out in the CHANGELOG); converting them would require env-aware rewriting the settled decision opted against.

The down migration is best-effort: it restores relative rows to the shipped Docker absolute form (the original spelling of each row is not recoverable after prefix collapsing).

## Test plan

- `uv run pytest tests/` — 500 passed; `ruff check .` and `uv run python -m mypy .` clean.
- New/updated tests:
  - `tests/test_podcast_audio_containment.py` — read helper containment (relative ok; `..` escape, absolute, `file://`, sibling-prefix, symlink escape, root itself all rejected) plus endpoint behavior (stream 403 for legacy/traversal, 404 for missing-in-root, 200 for valid relative; list/get omit `audio_url` for invalid values and set it for valid ones; delete refuses invalid unlink, unlinks valid relative).
  - `tests/test_podcast_audio_paths.py` — write helper (absolute/`file://`/CWD-relative under root → relative; outside root/traversal/root itself → `ValueError`; round-trip with the read helper) and migration registration/content checks (files exist, `AsyncMigrationManager` registers 21 up + down, prefix list and slice lengths match, single-strip invariant).
  - `tests/test_podcast_speaker_profile.py` — the end-to-end command test now asserts the stored/reported audio path is relative.
- **Migration validated against a live SurrealDB** (scratch namespace, then removed): 13 representative rows covering every converted prefix, the double-prefix pathological case (`/app/data/podcasts/data/podcasts/ep.mp3` → single strip only), `file://localhost/...` and percent-encoded URIs (left untouched), out-of-root absolutes (`/etc/passwd`, custom checkout paths — untouched), `NONE`, and already-relative rows. Down migration round-trips the converted rows to the Docker absolute form.

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1111?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lfnovo:
Reviewed — approve. The security core is solid and the refactor is complete.

## Containment (the part that matters) ✅

- `resolve_contained_audio_path()` is the single read-side choke point and it's correct: rejects empty, any `://` scheme, absolute paths, then `os.path.realpath(root / candidate)` + `is_relative_to(root)` catches `..` traversal and symlink-escape, and the explicit `resolved == root` check rejects the root-itself case. Write-side `to_relative_audio_path()` raises `ValueError` (permanent job failure) on anything outside the root, so the DB can never store an escaping value.
- **No bypass left.** The old per-router `_resolve_audio_path`/`_is_audio_path_contained` helpers are gone; all five consumption points (stream, list, get, delete, retry — retry via the shared delete helper) resolve through `resolve_contained_audio_path`. The remaining `episode.audio_file` references in the router are presence checks, log lines, and response serialization of the stored relative string — none touch the filesystem without the choke point.
- Write-side assigns `episode.audio_file = to_relative_audio_path(raw)` (line ~327), so stored values are validated-relative or None.

## Migration 21 ✅

- Prefix-strip conversion is correct — I checked each `string::slice` length against its prefix (`file://`→7, `/app/data/podcasts/`→19, `/data/podcasts/`→15, `./data/podcasts/`→16, `data/podcasts/`→14), and the IF/ELSE chain strips exactly one prefix. Rows under unknown roots stay untouched by design and read back as legacy-invalid (403/`audio_url: null`), matching #1018. Down migration is a documented best-effort restore.
- Numbered correctly — 20 (#1110) is on main, so 21 is next.

## Tests ✅

Genuinely thorough: `..` escape, absolute-even-inside-root, `file://`, and the sibling-prefix case (`../podcasts_evil/secret.mp3`) that a naive string-prefix containment check would miss; plus write-side ValueError and the slice-length assertions.

## Minor — align the registration test with the existing convention (cubic P3)

`test_manager_registers_migration_21` asserts `len(manager.up_migrations) == 21`, which the next migration (22) will break for an unrelated podcast-audio test. The migration-19 and migration-20 tests already established the tolerant pattern:

```python
assert len(manager.up_migrations) >= 21
assert len(manager.up_migrations) == len(manager.down_migrations)
assert "<distinctive SQL>" in manager.up_migrations[20].sql
```

Worth switching to that so a future migration doesn't trip this file. Non-blocking.

Fix the assertion and this is good to merge.

lfnovo:
Done — switched the registration test to the tolerant pattern from the migration-19/20 tests: `>= 21`, up/down parity, and a distinctive-SQL check on `up_migrations[20]`. Full file re-run green.

Issue #584, opened 2026-02-16
feat: migrate EpisodeProfile/SpeakerProfile to reference Model records for credential support


lfnovo:
## Problem

Podcast generation fails when users configure AI credentials only through the GUI (introduced in 1.7.0). The root cause is that `EpisodeProfile` and `SpeakerProfile` store raw provider/model name strings, which bypass the credential system entirely. Esperanto falls back to environment variables, which may not be set.

Related: #570
Upstream dependency: https://github.com/lfnovo/podcast-creator/issues/18

## Current State

**EpisodeProfile** stores 4 strings:
- `outline_provider` + `outline_model`
- `transcript_provider` + `transcript_model`

**SpeakerProfile** stores 2 strings:
- `tts_provider` + `tts_model`

These are passed directly to podcast-creator, which calls `AIFactory` with no credential config. Chat/Ask workflows work because they go through `ModelManager.get_model()` → `Credential` → `to_esperanto_config()`.

## Proposed Solution

Replace raw string pairs with references to **Model records**, which already carry provider, name, type, and an optional credential link. Use an additive + deprecation approach for safe migration:

### Phase 1: Schema changes (additive)

- Add new optional fields to `EpisodeProfile`: `outline_model_id`, `transcript_model_id` (references to `model` records)
- Add new optional field to `SpeakerProfile`: `tts_model_id` (reference to `model` record)
- Keep old string fields for backward compatibility
- Write SurrealDB migration to add the new fields

### Phase 2: Backend wiring

- Update `podcast_commands.py` to prefer Model record references when present, fall back to old strings
- Resolve Model → Credential → `to_esperanto_config()` and pass credential config to podcast-creator (requires https://github.com/lfnovo/podcast-creator/issues/18)
- Update API schemas to accept the new fields

### Phase 3: Frontend

- Update EpisodeProfile and SpeakerProfile forms to show Model record selectors (filtered by type: `language` for outline/transcript, `text_to_speech` for TTS)
- Populate new `*_model_id` fields when saving

### Phase 4: Deprecation (future release)

- Drop old string fields (`outline_provider`, `outline_model`, `transcript_provider`, `transcript_model` from EpisodeProfile; `tts_provider`, `tts_model` from SpeakerProfile)
- Migration to clean up any remaining profiles still using old fields

## Dependencies

- [ ] https://github.com/lfnovo/podcast-creator/issues/18 — podcast-creator must accept credential config dicts before open-notebook can pass them through

## Files affected

[...]

PR #1112, opened 2026-07-12
chore(podcasts): drop legacy provider/model string fields from podcast profiles

### LINKED ISSUE #1107: chore(podcasts): drop legacy provider/model string fields from podcast profiles (phase 4 of #584)
Phase 4 of #584 (phases 1-3 shipped in v1.11 — see that issue for history). The legacy provider/model string fields on podcast profiles are dead weight: the app ignores them (`open_notebook/podcasts/models.py` marks them "[Legacy]"), and their only remaining role is as input to the startup data migration's retry loop (`open_notebook/podcasts/migration.py`).

## Task

**One migration** (per ADR-006 — single release, no phasing), internal order:

1. **Best-effort mapping in SurrealQL**: for rows where the new reference fields (`outline_llm`, `transcript_llm` on `episode_profile`; `voice_model` on `speaker_profile`) are still empty, match legacy `provider` + `model` strings against existing `model` records and populate the references. (This mirrors what `podcasts/migration.py` does in Python, minus the auto-create path — no credential access in a migration.)
2. **Drop the 6 legacy fields**: `outline_provider`, `outline_model`, `transcript_provider`, `transcript_model` (episode_profile); `tts_provider`, `tts_model` (speaker_profile).
3. Remove the legacy fields from the Pydantic models and delete `open_notebook/podcasts/migration.py` + its startup hook (its job is done; the mapping now lives in the migration).

**Accepted trade-off (maintainer decision, 2026-07-12)**: profiles whose mapping never converged (no matching model record, e.g. credential never configured) lose the legacy strings and stay unresolved — they were already non-functional (the app ignores legacy fields, and the UI already flags them). The user re-picks models in the profile form once. No `legacy_config` stash. Needs a release-note entry stating this.

## Sequencing

Coordinate with #630 (speaker_config name→record reference) — same tables, and both add a migration. Do #630 first; this issue's migration takes the next number after it (ADR-006: numbers allocated in merge order).

## Acceptance criteria

- Fresh install: migration runs clean on empty tables
- Upgrade with mappable legacy rows: references populated, fields dropped, profiles work
- Upgrade with unmappable legacy rows: fields dropped, profile flagged in UI as needing model selection (existing behavior), no crash
- `podcasts/migration.py` and its startup call are gone; full suite passes
- Release note documents the trade-off

lfnovo:
## Summary

Phase 4 of #584 (phases 1-3 shipped in v1.11). The legacy provider/model string fields on podcast profiles (`outline_provider`, `outline_model`, `transcript_provider`, `transcript_model` on `episode_profile`; `tts_provider`, `tts_model` on `speaker_profile`) are removed end to end: database, Pydantic models, API schemas and frontend.

Closes #1107

## Migration notes (migration 22)

One migration per ADR-006, internal order:

1. **Best-effort mapping**: for rows where `outline_llm`/`transcript_llm`/`voice_model` are still empty but the legacy provider+model strings are set, populate the reference from a matching `model` record (`provider + name + type`, mirroring migration 20's `SELECT VALUE id FROM ONLY ... LIMIT 1` pattern). No auto-create — a migration must not touch credentials.
2. **`UNSET` the stored legacy values** while the fields are still defined, so SCHEMAFULL rows don't keep stale data the schema no longer knows.
3. **`REMOVE FIELD`** for all 6 legacy definitions.

The `_down` is a documented best effort: it restores the schema shape migration 14 left (`option<string>`), but the legacy strings themselves are not recoverable.

The startup data migration (`open_notebook/podcasts/migration.py`) and its `api/main.py` lifespan hook are deleted — the mapping now lives in the migration, and the auto-create-from-credential path is intentionally gone with it.

**Accepted trade-off** (maintainer decision in #1107): profiles whose mapping never converged (no matching model record, e.g. credential never configured) lose the legacy strings and stay unresolved — they were already non-functional (the app ignores legacy fields) and the UI already flags them as needing model selection; the user re-picks models in the profile form once. No `legacy_config` stash. The CHANGELOG entry documents this.

Frontend note: `PodcastEpisode.episode_profile`/`speaker_profile` are historical snapshots stored at generation time, so episodes generated before this change may still carry the legacy keys — those are typed via new `EpisodeProfileSnapshot`/`SpeakerProfileSnapshot` interfaces and `EpisodeCard` keeps rendering them; the live-profile interfaces and panels drop the legacy fallbacks.

## Test plan

- **Live migration validation** against a real SurrealDB (scratch namespace/database on the local 8018 instance) with seeded edge-case rows: mappable (reference populated, type filter verified against a same-name model of another type), unmappable (stays unresolved, no crash), already-migrated (reference NOT overwritten), fresh (untouched), empty-string legacy values (not mapped); verified no legacy values remain on any row and the field definitions are gone from `INFO FOR TABLE`; up runs clean on empty tables (fresh-install shape); down re-defines the fields; up re-runs after down.
- Also probed null-vs-NONE semantics: a bound Python `None` on these `option<...>` fields is stored as NONE (matched by the migration's `= NONE` guards) and a literal `null` is rejected by the schema, so no row can be skipped by the guard.
- New `tests/test_podcast_legacy_field_removal.py`: migration 22 registration (tolerant pattern), up/down content checks (map-before-drop ordering, UNSET-before-REMOVE, no auto-create), Pydantic fields gone, `open_notebook.podcasts.migration` import fails, lifespan hook gone.
- `uv run pytest tests/ -q`: 508 passed
- `ruff check .`: clean; `uv run python -m mypy .`: clean (122 files)
- Frontend: `npm run lint` (0 errors, 6 pre-existing warnings in untouched files), `npm run test` (116 passed), `npm run build` (success)

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1112?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lfnovo:
Reviewed — approve. Clean end-to-end removal, net −103 lines, and the migration is well-sequenced.

## Migration 22 ✅

- Correct ordering: best-effort map (`SELECT VALUE id FROM ONLY model WHERE provider+name+type LIMIT 1`, the validated migration-20 idiom that yields NONE on no match) → `UNSET` the stored legacy values while the fields still exist → `REMOVE FIELD IF EXISTS` for all 6. No auto-create (a migration shouldn't touch credentials) — unresolved profiles stay unresolved, the accepted #1107 trade-off.
- Down re-defines the 6 as `option<string>` (migration-14 shape), documented that the legacy string values themselves aren't recoverable.
- Numbered correctly — 21 (#1111) is on main.
- The registration test uses the tolerant `>= 22` + `up_migrations[21].sql` content assertion — nice, that's exactly the convention #1111 missed.

## Startup hook removal ✅

`open_notebook/podcasts/migration.py` is deleted and the `api/main.py` lifespan call is gone; the only remaining references are in the test that asserts the removal (import raises, `migrate_podcast_profiles` absent from main.py). No dangling production refs.

## Frontend ✅ — the snapshot typing is the right call

Legacy fields are dropped from the live `EpisodeProfile`/`SpeakerProfile` types, and `EpisodeProfileSnapshot`/`SpeakerProfileSnapshot` re-add them as optional historical fields. Since `PodcastEpisode.episode_profile` is a denormalized `model_dump()` snapshot (command line ~260), migration 22 — which touches the profile tables, not the stored episode dicts — leaves old episodes' snapshots intact. So `EpisodeCard` keeps compiling and old episodes still show their historical model info. Good backward-compat design; EpisodeCard correctly didn't need touching.

## Two cubic P2s — neither blocks

- **EpisodeCard shows `— / —` for new episodes**: valid, but not a regression — form-created profiles populate the reference fields (`outline_llm`), not the legacy strings, so new-episode snapshots were already empty here before this PR. The real improvement (render the resolved model from `outline_llm`/`transcript_llm`/`voice_model`) is a follow-up, out of scope for the field drop. Worth a small issue if we want those card rows informative again.
- **`extra="ignore"` accepts removed fields silently**: defensible — accept-and-discard is the more backward-compatible behavior for a field removal (old clients still sending `outline_provider` get a 200 instead of a new 422). `extra="forbid"` would make the removal loud at the cost of breaking old payloads. Trade-off, your call; I'd lean toward leaving it tolerant.

No changes requested. Good to merge.

lfnovo:
Filed the EpisodeCard display follow-up as #1114 (render the model rows from the `outline_llm`/`transcript_llm`/`voice_model` references instead of the dropped legacy fields). Tracked separately — doesn't block this merge.

Issue #1081, opened 2026-07-11
Add real-path tests for the core AI flows (chat graph, source ingestion) against a live SurrealDB in CI


lfnovo:
## Context

The July 2026 maintainability cleanup (PRs #1053–#1079) found that the test suite is strong on regression coverage but has **zero tests importing the core AI paths**: `open_notebook/graphs/chat`, `open_notebook/graphs/source` (source processing) and the end-to-end ingestion flow. Everything that exists for those paths is heavily mocked.

## Proposal

- Add a SurrealDB service container to the CI test workflow
- Add a small set of real-path tests: create source → process → embed → query; chat session → send message → verify graph output shape
- Keep them in a separate marker (e.g. `@pytest.mark.integration`) so the fast unit suite stays fast

## Why it matters

These are the flows users actually depend on; today a regression in them would ship silently if it doesn't trip a mocked regression test.

lfnovo:
## Moving to `needs-design` — one real fork to settle first

The infra pieces are standard, but there's a genuine design decision (how "real" the AI dimension is in CI) that should be settled before a dev picks this up.

### Current state (verified)

- CI `backend` job runs `uv run pytest tests/` with **no SurrealDB service**.
- **No pytest markers** are configured.
- Existing tests for these paths mock `repo_query` / the graph wholesale.

### Infra pieces (agreed regardless of the fork)

1. SurrealDB service container in the `backend` job of `.github/workflows/test.yml`, with migrations run against it.
2. A `@pytest.mark.integration` marker (registered in `pyproject.toml` `[tool.pytest.ini_options]`), run as a separate step so the fast unit suite stays fast.
3. Env wiring (`SURREAL_URL`, etc.) pointing at the service container.

### The fork — AI-provider strategy in CI (real API calls aren't an option)

esperanto gives three viable shapes:

- **(a) Local OpenAI-compatible stub** — a tiny fake server returning canned completions/embeddings; seed a credential with `base_url → localhost`. Exercises the full path (provision → esperanto → HTTP → parse → graph). Most faithful, deterministic, no keys. Cost: writing/maintaining the stub + seeding the credential in the test DB.
- **(b) Hybrid — real local embeddings + stubbed LLM.** esperanto's `transformers` embedding provider runs offline, so ingestion/embed/query is genuinely real; only the chat/transform LLM output is faked. Best signal-to-cost, no keys.
[...]

PR #1115, opened 2026-07-12
fix(podcasts): render EpisodeCard model details from resolved model references

### LINKED ISSUE #1114: EpisodeCard shows "— / —" for model details on new episodes (render from model references, not dropped legacy fields)
## Context

Follow-up from #1112 (Phase 4 of #584 / closes #1107), which dropped the legacy provider/model string fields from podcast profiles.

`EpisodeCard.tsx` (summary tab) still renders the model details from those now-removed legacy snapshot fields:

```tsx
// outline
{episode.episode_profile?.outline_provider ?? '—'} / {episode.episode_profile?.outline_model ?? '—'}
// transcript
{episode.episode_profile?.transcript_provider ?? '—'} / {episode.episode_profile?.transcript_model ?? '—'}
// speaker / tts
{episode.speaker_profile?.tts_provider ?? '—'} / {episode.speaker_profile?.tts_model ?? '—'}
```

Since #1112, newly generated episodes' denormalized `episode_profile`/`speaker_profile` snapshots no longer carry those fields (they were only ever populated for very old rows, if at all), so these three rows now render **"— / —"** for every new episode. The `EpisodeProfileSnapshot`/`SpeakerProfileSnapshot` types keep the fields optional purely so `EpisodeCard` still compiles and *old* episodes keep showing their historical values — this is intentional backward-compat, not the intended long-term display.

Note: this is **not a regression** from #1112 — form-created profiles use the Model-registry references (`outline_llm`, `transcript_llm`, `voice_model`), not the legacy strings, so new-episode cards were already showing "—" before the drop. #1112 just makes it permanent.

## Proposal

Render the model details in `EpisodeCard` from the reference fields the snapshot still carries — `episode_profile.outline_llm`, `episode_profile.transcript_llm`, `speaker_profile.voice_model` (model record IDs) — resolved to a human-readable model name.

Options for the resolution:
- **Frontend-side**: resolve the IDs against the models list the settings UI already fetches (a `useModels()`-style hook / react-query cache), mapping record ID → `provider/name`.
- **Backend-side**: add resolved display fields to the episode/snapshot serialization (e.g. `outline_model_name`) so the card doesn't need to resolve IDs itself — mirrors how #1110 added `speaker_config_name` alongside the `speaker_config` record ID.

The backend-resolved approach is probably cleaner and consistent with the `speaker_config_name` precedent, but either works.

## Acceptance criteria

- `EpisodeCard`'s Outline / Transcript / Speaker model rows show the resolved model (provider + name, or a sensible label) for newly generated episodes instead of "— / —".
- Old episodes that only have the legacy snapshot strings still render something reasonable (fall back to the historical strings, or the resolved reference if present).
- Unresolvable references (deleted model) degrade gracefully to "—".

## Scope note

Cosmetic / display-only, not user-blocking — the profiles themselves work through the Model registry references. No rush.

lfnovo:
Closes #1114 (last item, cosmetic). Since #1112 dropped the legacy provider/model strings, `EpisodeCard`'s outline / transcript / speaker rows rendered `— / —` for new episodes — the snapshots carry `outline_llm`/`transcript_llm`/`voice_model` record IDs instead.

## Approach
Backend-side resolution, consistent with the `speaker_config_name` precedent from #630:
- New `Model.get_display_info_for_ids()` batch-fetches `{provider, name}` for many model IDs in **one query** (`SELECT ... WHERE id IN $model_ids`).
- `list_podcast_episodes` collects the distinct references across the whole page and resolves them once — **no per-row lookup** (the no-N+1 acceptance criterion). `get_podcast_episode` does the same for the single episode.
- `EpisodeCard`'s `formatModelLabel` renders resolved provider/name, falling back to the legacy snapshot strings (old episodes), then to `—` (unresolvable / deleted model).

## Test plan
- Backend: `tests/test_podcast_episode_model_resolution.py` — resolved / legacy-only / unresolvable episodes, single-query batching, dedup+falsy filtering, query-failure-degrades-to-empty, and an assertion that the batch method is called once with no per-episode `Model.get`. Full suite 520 passed, ruff + mypy clean.
- Frontend: `EpisodeCard.test.tsx` covers the three render cases. lint 0 errors, tsc clean, vitest 119 passed, build green.

## Note for review
This PR was implemented by a worker that was interrupted before its self-review; I picked it up, and in doing so fixed a mypy `arg-type` error (list invariance on `get_display_info_for_ids` → `Sequence`) and a resulting unused-import that the interrupted run left behind. **One thing worth a focused look**: the resolution tests mock `repo_query` with string `id`s, so they exercise the logic but not the real `RecordID` → `str` matching against the snapshot's stored reference form. It mirrors the battle-tested `get_job_details_for_commands` batching pattern and relies on the same `str(RecordID) == 'table:id'` behavior used across the codebase, but it isn't validated against a live DB here.

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1115?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lfnovo:
Reviewed — approve. Cleanly closes #1114 with the backend-resolution approach, and the no-N+1 requirement is genuinely met.

## Backend ✅

- `Model.get_display_info_for_ids()` batches correctly: `sorted({str(mid) for mid in model_ids if mid})` dedups + drops falsy, one `SELECT ... WHERE id IN $model_ids` for the whole set, and a query failure returns `{}` so display resolution degrades to no-fields rather than failing the request.
- `list_podcast_episodes` collects the distinct refs across the whole page (`_collect_snapshot_model_ids`) and resolves once — no per-row lookup — and `get_podcast_episode` does the single-episode equivalent. Mirrors the `get_job_details_for_commands` batching precedent.
- **The RecordID→str matching is sound**: the grouped dict is keyed by `str(row.get("id"))` (line 101) and the caller looks up `models_by_id.get(str(ref))` (both sides `str()`), and the snapshot stores `outline_llm`/`transcript_llm`/`voice_model` as `Optional[str]` (`model_dump()` → `"model:xxx"`), so the keys line up. On your review note: I couldn't exercise the live round-trip (the dev DB has no `model` rows), but the keying is symmetric and this is the same `str(RecordID) == "table:id"` assumption already load-bearing in `get_job_details_for_commands`, so the risk is low — good call flagging it, and a live-DB integration test (per #1081) would close it for good.

## Frontend ✅

- `formatModelLabel(provider, name, legacyProvider, legacyName)` → `${provider || legacyProvider || '—'} / ${name || legacyName || '—'}` is exactly the resolved → legacy-snapshot → `—` chain #1114 asked for, applied to all three rows; the new `*_model_provider`/`*_model_name` types match the API serialization.

## Tests ✅

Backend covers resolved / legacy-only / unresolvable, single-query batching, dedup + falsy filtering, and failure-degrades-to-empty; frontend covers the three render cases including `— / —` ×3. Matches #1114's acceptance criteria point for point.

Nice cleanup of the interrupted-worker leftovers (the `Sequence` invariance fix + unused import). No changes requested — good to merge.

Issue #939, opened 2026-06-21
[Tech] Upgrade to content-core 2.x (Reddit, MP3 chapters, Crawl4AI, pre-flight validation, MIT licensing)


lfnovo:
## Summary

Open Notebook pins `content-core>=1.14.1,<2` (`pyproject.toml:40`). content-core **2.x** is a *"Major Architecture Refactor"* that fixes several issues we can't pick up on the 1.x line and ships new capabilities we want to surface. Target: bump to **content-core `>=2.0.4,<3`** and adapt our single integration point (`open_notebook/graphs/source.py`).

This umbrella now carries the full design (the *how*), not just the tracking list. It's scoped in workstreams so the mandatory migration can ship first and the opt-in features can land in the same release or immediately after.

## Children (tracked)

- [ ] **#1103** — core migration (bump + `source.py`) — **foundation, blocks all**; verifies & closes #480, #698
- [ ] **#975** — pre-flight `check_file_support()` for unsupported uploads
- [ ] **#432** — Crawl4AI as a URL engine (+ Docker mode)
- [ ] **#602** — Firecrawl `proxy` / `wait_for` options
- [ ] **#1104** — Docling OCR toggle (scanned PDFs / images)
- [ ] **#1105** — docs: engines page + supported formats + ADR-002

Verified-and-closed-by #1103 (no separate work): #480 (Reddit), #698 (MP3 chapters).

---

## What the upgrade gives us

**Fixed for free (verify end-to-end):**
- **#480** — Reddit URLs (403 anti-bot). Now extracted via Reddit's public JSON endpoint, with fallback to normal URL extraction.
- **#698** — ffmpeg/moviepy crash on MP3 files with chapter metadata. moviepy replaced with direct ffmpeg/ffprobe calls.
- YouTube `youtube.com/live/<id>` and `/shorts/<id>` URLs now recognized (2.0.4) — previously failed title + transcript extraction. No ON issue, but worth a regression test.
- **License hygiene**: PyMuPDF (**AGPL3**) → pdfplumber (MIT); moviepy removed. Relevant for a self-hostable project — call this out in the release notes and in `ADR-002`.

**New capabilities to surface (opt-in, need UI + config):**
- **#432** — Crawl4AI as a URL engine (+ Docker mode via `CRAWL4AI_API_URL`).
- **#602** — Firecrawl `proxy` (default `auto`) and `wait_for` (default 3000ms) to bypass reCAPTCHA/anti-bot.
[...]

PR #1117, opened 2026-07-12
fix(sources): reject unsupported uploads at ingestion with 415 (#975)

### LINKED ISSUE #975: [Bug]: Unsupported file types trigger 15 retry attempts over ~1 hour before showing generic "Failed" with no actionable error

lfnovo:
Part of #939 — **Wave 2**. Builds on the content-core 2.x migration (#1116, merged), which shipped the `check_file_support()` API this uses.

## Problem

Uploading a file content-core can't extract (e.g. `.zip`, `.webp`) enqueued a background `process_source` job that failed, then retried **up to 15 times over ~1 hour** (`stop_on=[ValueError]` doesn't stop a content-core `UnsupportedTypeException`), before finally surfacing a generic "Failed" with no actionable detail (#975).

## Fix

- **Pre-flight validation** — a new `_assert_file_supported()` calls content-core 2.x's `check_file_support(file_path)`, which does header-only identification + routing using the *same* logic real extraction uses (so the verdict can't disagree). It runs in the upload branch of `_build_content_state`, **before** the source record is created and **before** any job is enqueued, so an unsupported upload never reaches the worker.
- **Retry endpoint guarded** the same way, so an existing failed source of an unsupported type can't be re-queued into the same storm.
- **`415 Unsupported Media Type`** — added a dedicated global handler for `UnsupportedTypeException` (it previously fell through to the base `OpenNotebookError` handler → 500). The response names the detected MIME type.
- **Defensive** — if `check_file_support()` itself errors (e.g. the file was removed before a retry), the check is skipped and normal extraction handles it, so a transient issue never becomes a hard 415.

The rejected upload is cleaned off disk by `create_source`'s existing `except OpenNotebookError` arm (no orphaned files).

## Testing

- `uv run pytest tests/` — 527 passed (+4). New: `_assert_file_supported` unit cases (unsupported → raises, supported → passes, check-error → swallowed) in `tests/test_upload_type_mitigations.py`, and a `UnsupportedTypeException → 415` handler-mapping case in `tests/test_typed_exceptions_reach_handlers.py`.
- `ruff check .` and `uv run python -m mypy .` — clean.
- Runtime-checked real `check_file_support()` verdicts: random bytes → `supported=False` with a reason; a markdown file → `supported=True` (`text` processor).

Closes #975

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1117?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lfnovo:
## Review — reject unsupported uploads with 415 at ingestion ✅ approve

Wave-2 of #939, closes #975. Reviewed manually + cubic CLI (clean, `{"issues": []}`); verified the content-core surface against the installed `2.0.4`.

### Verified correct
- **`check_file_support()` contract** — runtime-checked: a `.zip` of random bytes → `supported=False`, `reason="Unsupported file type: application/zip"`, `identified_type="application/zip"`; a markdown file → `supported=True` (`text/plain`). `_assert_file_supported` consumes exactly those fields, and the message correctly appends the detected type for the UI.
- **Same routing as real extraction** — `check_file_support` runs content-core's header-only `get_file_type` → detector, the same identification path `extract_content` uses, so the pre-flight verdict can't disagree with what the worker would do.
- **415 handler** — `UnsupportedTypeException` is Open Notebook's **own** class (`open_notebook.exceptions`, subclass of `OpenNotebookError`), distinct from content-core's. The dedicated `@app.exception_handler(UnsupportedTypeException)` resolves ahead of the base `OpenNotebookError` handler via MRO, so the pre-flight raise maps to 415, not the old 500.
- **No orphaned files** — the raise happens inside `create_source`'s `try`; `UnsupportedTypeException` (an `OpenNotebookError`, not `InvalidInputError`) lands in the `except OpenNotebookError` arm → `_cleanup_uploaded_file` → re-raise → 415. File is removed.
- **Placement** — pre-flight runs *after* the uploads-dir path-traversal guard and *before* the record/job are created; the **retry** endpoint is guarded identically, so a stored unsupported source can't be re-queued into the storm.
- **Defensive swallow** — a `check_file_support` error (e.g. file gone on retry) is logged and skipped rather than becoming a hard 415.
- **Tests** — green here (26 passed): `_assert_file_supported` raises/passes/swallow, plus a handler-mapping test that injects `UnsupportedTypeException` into an endpoint and asserts 415 + type in the detail.

### One residual (non-blocking, optional)
The defensive swallow leaves a narrow window: if `check_file_support()` itself throws and is swallowed, the file still enqueues, and in the worker content-core's *own* `UnsupportedTypeException` isn't a `ValueError`, so `stop_on=[ValueError]` won't stop the 15× retry — the original #975 behavior, only in that rare check-error corner. Could be hardened later by mapping content-core's `UnsupportedTypeException` → `ValueError` in `content_process` so `stop_on` catches it. Not needed for this PR. LGTM to merge.

Issue #1105, opened 2026-07-12
[Docs] Content processing engines page + supported formats refresh + ADR-002 update


lfnovo:
Part of #939 (content-core 2.x upgrade). Cross-cutting documentation — pairs with the feature children (#432, #602, OCR toggle) but tracked separately because it spans the whole upgrade.

## Goal

After the 2.x upgrade, users should understand **what source formats/engines are available** and **how to configure them**, and our decision log should reflect the major bump.

## Tasks

1. **New page: "Content processing engines"** — there is no page today explaining the doc/url engines and when to pick each:
   - URL engines: `auto` (Firecrawl→jina→crawl4ai→bs4 fallback), `firecrawl`, `jina`, `crawl4ai`, `simple` (bs4).
   - Doc engines: `auto`, `docling`, `simple`.
   - Note: `docs/7-DEVELOPMENT/content-processing.md` is misleadingly named — it covers chunking/embedding, not engines. Either add the new page under user docs or clearly separate the two.

2. **Refresh `docs/3-USER-GUIDE/adding-sources.md`**:
   - Supported formats list (EPUB now supported; Reddit URLs; YouTube live/shorts; OCR note for scanned PDFs/images).
   - "Unsupported file type" troubleshooting (L347-360) — reflect the new early-rejection behavior with an actionable message (see #975), instead of the retry storm.

3. **Config docs** — `docs/5-CONFIGURATION/environment-reference.md` + `advanced.md`: document `CRAWL4AI_API_URL` (Crawl4AI Docker mode, #432), `FIRECRAWL_API_URL` / `FIRECRAWL_API_KEY`, and any `CCORE_*` knobs we rely on.

4. **`docs/7-DEVELOPMENT/decisions/ADR-002-external-libraries.md`** — update for the content-core 2.x major bump and the **license change** (PyMuPDF AGPL3 → pdfplumber MIT; moviepy removed). A short decision-log note that we moved to 2.x.

## Done when
- A user can find, from the docs, which formats are supported and how to switch/configure engines.
- ADR-002 reflects the 2.x decision and the licensing improvement.

Issue #1104, opened 2026-07-12
[Feature] Expose Docling OCR toggle for scanned PDFs and image sources


lfnovo:
Part of #939 (content-core 2.x upgrade). Depends on the core migration child of #939.

## Goal

content-core 2.x exposes Docling enrichment flags on `ContentCoreConfig`: `docling_ocr`, `docling_formulas`, `docling_vision`. Surface at least **OCR** so scanned PDFs and image sources (JPG/PNG/TIFF) get text extracted, instead of silently coming back empty. Connects to #331 (multimodal strategy) and #775 (OCR on watermarked PDFs).

## Scope (vertical slice — backend + UI + i18n + tests + docs)

**Backend**
- `open_notebook/domain/content_settings.py` — add settings field(s): `docling_ocr` (bool, at minimum). Decide whether to also expose `docling_formulas` / `docling_vision` now or defer to #331.
- `open_notebook/graphs/source.py` — pass the flag(s) through `ContentCoreConfig`.
- `api/models.py` + `api/routers/settings.py` — extend the settings request/response + validation.

**UI**
- `frontend/src/app/(dashboard)/settings/components/SettingsForm.tsx` — add an OCR toggle in the "Content Processing" card with helper text ("extract text from scanned PDFs and images; slower").

**i18n (mandatory)**
- All 15 locale dirs under `frontend/src/lib/locales/*/index.ts` — labels + help text. English-only blocks the release.

**Tests**
- Domain default for the new field (`tests/test_domain.py`).
- A source-processing test asserting the flag reaches `ContentCoreConfig`.

**Docs**
- Mention OCR in the supported-formats/troubleshooting docs (coordinated with the docs child of #939).

## Notes / decisions
- Docling OCR is meaningfully slower and needs the docling extra installed — confirm it's available in the Docker image before defaulting it on. Recommend defaulting **off** with a clear toggle.
- `DOCLING_SUPPORTED` in content-core already includes `image/png`, `image/jpeg`, `image/tiff`, `image/bmp` — so image OCR routes through docling when the engine allows it.

Issue #602, opened 2026-02-18
[Feature]: Add Firecrawl proxy and waitFor options to bypass reCAPTCHA/anti-bot protection


vovasuglob:
## Problem

When using Firecrawl as the URL extraction engine, many websites block scraping with reCAPTCHA or other anti-bot protections. This causes content extraction to fail, which breaks podcast generation and other features that depend on web content.

The Firecrawl API (v2) supports `proxy` and `waitFor` parameters in `ScrapeOptions` that can help bypass these protections:
- `proxy: "auto"` — automatically retries with stealth proxies if the basic request is blocked
- `wait_for: 3000` — waits for dynamic content/challenges to resolve before extracting

However, the current `_fetch_url_firecrawl` function in `content_core/processors/url.py` does not pass these parameters:

# Current code (line ~197)
scrape_result = await app.scrape(url, formats=["markdown", "html"])

### Why would this be helpful?

# Suggested change
from firecrawl.v2.types import ScrapeOptions

proxy_mode = os.environ.get("FIRECRAWL_PROXY", "auto")  # basic, stealth, auto
wait_for_ms = int(os.environ.get("FIRECRAWL_WAIT_FOR", "3000"))

scrape_options = ScrapeOptions(
    formats=["markdown", "html"],
    proxy=proxy_mode,
    wait_for=wait_for_ms,
)
scrape_result = await app.scrape(url, options=scrape_options)

### Contribution

- [ ] I am a developer and would like to work on implementing this feature (pending maintainer approval)

lfnovo:
This is coming soon

lfnovo:
Great suggestion with a clear fix! This lives in content-core's URL processor. Tracked upstream at lfnovo/content-core#34.

lfnovo:
Tracked under the content-core 2.x upgrade umbrella #939. Depends on the core migration (#1103) landing first.

lfnovo:
Triage update: removed `ready` — this is upstream work, tracked in content-core.

The upstream issue lfnovo/content-core#34 was previously closed as completed, but on checking the shipped **content-core 2.0.4** (what we pin), the Firecrawl scrape call still passes no `proxy` / `wait_for` — the fix isn't actually in. I've **reopened content-core#34** with that evidence: lfnovo/content-core#34.

[...]

PR #1118, opened 2026-07-12
feat(sources): add Crawl4AI URL engine + honor persisted engine settings (#432)

### LINKED ISSUE #432: [Feature]: Enable Crawl4AI option
### Feature Description

Content-core now supports Crawl4AI as of 1.9.0
We should add this option to the project

Should help people do extractions without needing API Keys

ref: https://github.com/lfnovo/content-core/pull/21
https://github.com/lfnovo/content-core/releases/tag/v1.9.0

_No response_

### Additional Context

_No response_

### Contribution

lfnovo:
Part of #939 — **Wave 2**. Builds on the content-core 2.x migration (#1116, merged), which added the Crawl4AI URL processor.

## What this adds

- **Crawl4AI as a URL engine** — selectable in Settings → Content Processing next to Firecrawl / Jina / Simple. It renders JavaScript-heavy pages locally (Playwright) with no API key, or offloads to a remote Crawl4AI server when `CRAWL4AI_API_URL` is set (content-core reads that env var natively).
- **Persisted engine settings now actually take effect.** The source-processing graph previously built a *hard-coded* `ContentSettings(...)` with `auto` engines and never read the saved value — so the doc/URL engine selectors in Settings were dead UI. The graph now loads `ContentSettings.get_instance()` and passes both engines to `ContentCoreConfig`, falling back to defaults if settings can't load. This is what makes #432 (and the upcoming #602/#1104) reachable at all.

## Changes

- **Backend**: `crawl4ai` added to the URL-engine `Literal` (`content_settings.py`) and the settings-update `cast` (`settings.py`); `graphs/source.py` loads persisted settings into `ContentCoreConfig`.
- **Frontend**: zod enum + `<SelectItem value="crawl4ai">` + reset cast in `SettingsForm.tsx`; `crawl4ai` label added to **all 14 locales**; en-US `urlHelp` updated to describe the new `auto` fallback chain (Firecrawl → Jina → Crawl4AI → simple).

## Testing

- `uv run pytest tests/` — 525 passed. New: `ContentSettings` accepts `crawl4ai` (`test_domain.py`); the persisted engines reach `ContentCoreConfig` (`test_graphs.py`).
- Backend `ruff` + `mypy` clean.
- Frontend `npm run lint`, `npm run build`, and the i18n parity test (`src/lib/locales`) all pass — every locale has the new key (`satisfies TranslationShape`).

## Notes

- Crawl4AI env/config documentation (`CRAWL4AI_API_URL`) is deferred to the docs child #1105.
- Depends on nothing else in Wave 2; #1104 (OCR toggle) will build on the same settings-loading path, so it should merge after this to avoid churn.

Closes #432

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1118?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lfnovo:
## Review — Crawl4AI URL engine + honor persisted engine settings ✅ approve (one change requested before merge)

Wave-2 of #939, closes #432. Reviewed manually + cubic CLI (clean, `{"issues": []}`); verified the content-core surface against the installed `2.0.4`.

### Verified correct
- **The real fix works.** The graph previously built a hard-coded `ContentSettings(...)` and never read the saved value, so the doc/URL engine selectors were dead UI. `content_process` now loads `ContentSettings.get_instance()` and passes both engines into `ContentCoreConfig`, falling back to defaults on load failure. Confirmed via `test_persisted_engines_wired_into_config` (asserts `config.url_engine == "crawl4ai"` / `document_engine == "docling"` reach `extract_content`).
- **content-core supports crawl4ai** — `VALID_URL_ENGINES = {"auto","simple","firecrawl","jina","crawl4ai"}`, dedicated `processors/url/crawl4ai.py` (local Playwright + remote Docker via `CRAWL4AI_API_URL`), and `_extract_url_with_engine` routes `engine == "crawl4ai"` → `extract_url_crawl4ai`.
- **`auto` mode stays safe.** `extract_url_crawl4ai` swallows failures (incl. the not-installed `ImportError`) and returns `None`, so the `auto` chain (Firecrawl → Jina → Crawl4AI → bs4) correctly falls through to BeautifulSoup. No regression for existing `auto` users.
- **Frontend** — zod enum, reset cast, and `<SelectItem value="crawl4ai">` all consistent; the `crawl4ai` label lands in all 14 locales (i18n parity test enforces it).
- **Tests** — green here (48 passed): the Literal accepts `crawl4ai` (`test_domain.py`) and the persisted engines reach the config (`test_graphs.py`).

### 🔧 Change requested before merge — bundle the Crawl4AI runtime
The **`crawl4ai` package is not a dependency of Open Notebook** — it's not in `pyproject.toml`, the `Dockerfile`, or `uv.lock` (0 occurrences); content-core declares it only as an optional `content-core[crawl4ai]` extra. So in a stock install (Docker or pip) with no `CRAWL4AI_API_URL`:
- selecting **Crawl4AI** → `extract_url_crawl4ai` returns `None` → empty extraction → clean `ValueError` job failure. Every URL fails.
- and the en-US `urlHelp` promises *"Crawl4AI renders JavaScript pages locally … with no API key"*, which isn't true out of the box.

Decision (with the maintainer): **make local mode work by default.** Please bundle it into the default image so the option is honest:
1. `pyproject.toml`: `content-core>=2.0.4,<3` → `content-core[crawl4ai]>=2.0.4,<3`, then `uv lock`. Footprint is modest — 27 packages, **no torch/transformers/CUDA**; heaviest is `scipy`.
2. `Dockerfile` (`runtime-base`, so both the regular and single variants get it): after the venv is copied, install the Chromium runtime, e.g. `.venv/bin/playwright install --with-deps chromium` with a fixed `PLAYWRIGHT_BROWSERS_PATH` and apt-list cleanup. This adds ~300 MB (Chromium + system libs) to the image — accepted, as it's what makes the local no-API-key promise real.
3. Keep the `urlHelp` text as-is (it becomes accurate once bundled).

`auto` is safe and the wiring is correct, so this is purely about making the newly-exposed local engine actually usable. LGTM once the runtime is bundled.

lfnovo:
Addressed the review in fcc53d7.

**Change requested — Crawl4AI runtime bundled (decision: make local mode work by default):**
- `pyproject.toml` → `content-core[crawl4ai]>=2.0.4,<3`, `uv lock` refreshed. Confirmed the footprint is modest — **no torch/transformers/CUDA** pulled in.
- `Dockerfile` `runtime-base` (so both the `single` and `runtime` variants inherit it): installs the Chromium browser via `playwright install --with-deps chromium` after the venv is copied, pinned to `PLAYWRIGHT_BROWSERS_PATH=/app/pw-browsers`. Adds ~300 MB (Chromium + system libs).
- `urlHelp` kept as-is — it's now accurate (local no-API-key mode works out of the box).

**cubic findings:**
- ✅ `ContentSettings` singleton leak in tests → added `clear_instance()` teardown to `TestContentSettings`.
- ✅ Lost traceback on settings-load failure → now `logger.opt(exception=True).warning(...)`, preserving the server-side traceback while still falling back to defaults.
- **Non-en `urlHelp` not updated for Crawl4AI (P2/P3 ×~7):** intentionally not translating the Crawl4AI blurb into 13 locales in this PR — the label is present everywhere (parity holds), and the "what is each engine" copy is being consolidated into the new engines docs page under #1105 rather than duplicated as prose across 14 locale files. The en-US help is updated for reference.

All green after the change: 525 backend tests, `ruff`/`mypy` clean, frontend lint/build + i18n parity pass.

lfnovo:
## Re-review of `fcc53d7` (bundling addressed)

The requested Crawl4AI runtime bundling is in and verified:
- `pyproject.toml` → `content-core[crawl4ai]`; `uv.lock` refreshed and consistent (`uv lock --check` clean, so the Docker `uv sync --frozen` won't break).
- `Dockerfile` `runtime-base`: `playwright install --with-deps chromium` with a pinned `PLAYWRIGHT_BROWSERS_PATH=/app/pw-browsers`. **`build-regular` passed (5m53s)**, so the browser + system-lib install works on `python:3.12-slim-trixie` (including the arm64/QEMU leg). The browser path survives the later `COPY . /app` and the `/app/data` volume mount.
- Nice extras: `logger.opt(exception=True)` preserves the traceback on the settings-load fallback, and the new `teardown_method` + `ContentSettings.clear_instance()` isolates the `ContentSettings` singleton across tests. Affected tests green; all CI green.

### cubic flagged one design point (P2) — surfacing for the author to decide
> Making `content-core[crawl4ai]` a **base** dependency turns the optional URL engine into a mandatory runtime surface for every install, pulling the crawling stack's transitive Python deps even for users on Simple/Jina/Firecrawl or a remote Crawl4AI server.

Neutral notes to inform the call:
- The heavy part (Chromium, ~170 MB) is **Docker-only** via the explicit `playwright install` step — a bare `uv sync` pulls only the ~27 Python packages (no torch/CUDA; heaviest is `scipy`), not the browser.
- Open Notebook is Docker-first (README/docs promote Docker Compose + from-source; no `pip install open-notebook` user path), so the base-dependency surface is mainly felt by from-source contributors.
- Alternative if you want the default install kept lean: expose an Open Notebook extra — `[project.optional-dependencies] crawl4ai = ["content-core[crawl4ai]"]` — and have the Dockerfile opt in with `uv sync --extra crawl4ai`. Same Docker-out-of-the-box result, leaner from-source install, at the cost of a bit more pyproject/Dockerfile wiring.

No recommendation attached — your call on base-dep vs. extra. Also note the branch is currently behind `main` and will need an update before merge.

lfnovo:
Re: the re-review — the two code findings were already addressed in fcc53d7; cubic re-posted them against the pre-fix commit:

- **`ContentSettings` singleton leak** — fixed: `TestContentSettings` now has a `teardown_method` calling `ContentSettings.clear_instance()` (covers every test in the class, not just the crawl4ai one).
- **Lost traceback on settings load** — fixed: now `logger.opt(exception=True).warning(...)`, which keeps the full server-side traceback while staying at WARNING (non-fatal, falls back to defaults).

**Non-en `urlHelp` Crawl4AI drift (P2/P3 ×7)** — unchanged by design. The engine *label* is present in all 14 locales (i18n parity holds; the option is fully usable everywhere). We're deliberately **not** duplicating the "what each engine does" prose across 14 locale files — that copy is being consolidated into the new content-processing-engines docs page under #1105. en-US keeps the reference description. Happy to revert the en-US `urlHelp` addition instead if you'd prefer all locales identical — let me know.

PR #1121, opened 2026-07-12
docs: content processing engines + extraction config (#1105, #602)

lfnovo:
Part of #939 — the final child. Documents everything the content-core 2.x upgrade exposed, and closes **#602** (Firecrawl proxy/wait_for — env-only, so it was always a docs task).

## What's here

- **New page** `docs/3-USER-GUIDE/content-processing-engines.md` — explains the document engines (`auto`/`docling`/`simple`) and URL engines (`auto`/`firecrawl`/`jina`/`crawl4ai`/`simple`), how the `auto` fallback chain works (Firecrawl → Jina → Crawl4AI → simple), the OCR toggle, and a troubleshooting quick-reference. Linked from the User Guide index.
- **`adding-sources.md`** — image/OCR support, Reddit URLs, YouTube `/live/` + `/shorts/`, JavaScript-heavy-site guidance, and the new **immediate 415 rejection** for unsupported types (no more long "Processing" hang, per #975).
- **`environment-reference.md`** + **`advanced.md`** — `FIRECRAWL_API_URL`, `CCORE_FIRECRAWL_PROXY`, `CCORE_FIRECRAWL_WAIT_FOR`, `CRAWL4AI_API_URL`, with a note that `CCORE_*` vars pass straight through to content-core. **This is the resolution of #602** (no code needed — content-core reads them natively).
- **`ADR-002`** — addendum on the 2.x upgrade and the AGPL→MIT licensing improvement (PyMuPDF → pdfplumber, moviepy → ffmpeg).

## Notes

- Docs are English-only (no i18n). There's no MkDocs/Docusaurus nav file — navigation is via per-section `index.md`, so the new page is registered as a "Related" link under Adding Sources in `docs/3-USER-GUIDE/index.md`.
- Verified with the repo's own `scripts/check_md_links.py` — all relative links resolve.

With this merged, the #939 umbrella (content-core 2.x) is complete.

Closes #1105
Closes #602

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1121?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Issue #374, opened 2026-01-01
[Umbrella] Offline & Air-Gapped Deployment


lfnovo:
# 🔌 Offline & Air-Gapped Deployment

This umbrella tracks work to make Open Notebook fully functional without internet access — critical for privacy-focused users and secure environments.

## Why This Matters

Many users choose Open Notebook specifically for privacy and data sovereignty:
- **Security-conscious organizations** running in isolated networks
- **Researchers** working with sensitive data
- **Users in low-connectivity areas** who need reliable local operation
- **Privacy advocates** who want zero external calls

Currently, Open Notebook requires internet access for certain operations. This umbrella tracks fixing that.

### 1. Tiktoken Downloads (~Easy)
**Issue**: The `tiktoken` library downloads tokenizer files on first use.

**Solution**: Pre-cache tokenizer files in the Docker image.

```python
# During build, pre-download:
import tiktoken
tiktoken.get_encoding("cl100k_base")
tiktoken.get_encoding("p50k_base")
# etc.
```

**Status**: Well understood, needs implementation.

### 2. Docker First-Boot Dependencies (~Medium)
**Issue**: Docker container tries to download Python packages on first startup (#265).

**Solution**: Ensure all dependencies are baked into the image, no runtime downloads.

**Investigation needed**: Why is `virtualenv` being downloaded at runtime?

### 3. Content Extraction (#264) (~Medium)
**Issue**: Content extraction may require external services.

**Proposed solution**: Support local extraction via Apache Tika or similar.

**Status**: Needs architecture decision — is this in scope for v1?

### Core
- [ ] #264 — Enable use without internet access (remove tiktoken dependency)
- [ ] #265 — Docker install with no internet connection

### Future (out of scope for now)
- [ ] Local content extraction (Apache Tika integration)

### Tiktoken Caching (Good First Issue)
**Difficulty**: Easy | **Skills**: Python, Docker

1. Identify all tiktoken encodings used by Open Notebook
2. Add pre-download step to Dockerfile
3. Test in air-gapped environment
4. Submit PR

### Docker Dependencies Investigation
**Difficulty**: Medium | **Skills**: Docker, Python packaging

1. Reproduce the issue from #265
2. Identify what triggers runtime downloads
3. Fix the Dockerfile to include all dependencies
4. Test in isolated network
5. Submit PR

## Getting Started

```bash
# Clone
git clone https://github.com/lfnovo/open-notebook
cd open-notebook

# Test current offline behavior
[...]

PR #1123, opened 2026-07-13
feat(sources): opt-in Docling + Crawl4AI runtimes installed at startup (#1122)

### LINKED ISSUE #1105: [Docs] Content processing engines page + supported formats refresh + ADR-002 update

### LINKED ISSUE #1122: [Tech] Opt-in heavy extraction runtimes (Docling OCR + Crawl4AI local) via env-gated startup install
## Summary

Make the two **heavy** extraction runtimes — **Docling** (document layout + OCR) and **Crawl4AI local mode** (JavaScript rendering via Chromium) — **opt-in**, installed at container startup when the operator enables them via an env var, with downloads cached on the data volume. Keep the default image lean.

This reverses the "bundle Crawl4AI by default" decision from #432 (which added ~300 MB of Chromium to every image) and makes the OCR toggle from #1104 actually functional (today it's a silent no-op because Docling isn't installed). It also avoids ever baking Docling's multi-GB ML stack (torch + transformers + peft + models) into the default image.

Inspired by the ComfyUI / Automatic1111-in-Docker pattern (entrypoint installs on first run, volume persists), but deliberately persisting **only the download caches** — not the venv — to avoid the Python-version-coupling fragility those projects live with.

## Background / problem

- `content-core` extras: `docling = ["docling>=2.86.0", "peft>=0.18.0", "transformers>=4.50,<5"]` (pulls torch transitively — multi-GB) and `crawl4ai = ["crawl4ai>=0.7.0"]` (needs a Chromium browser).
- **Docling is not installed today** (0 occurrences in `uv.lock`; we depend on `content-core[crawl4ai]`). So:
  - The **document engine `docling`** silently falls back (it has always been a selectable option that no-ops without the package).
  - The **OCR toggle (#1104)** is a no-op — OCR only runs through Docling.
  - **Image sources** (`.png/.jpg/.tiff/.bmp`) are unsupported (route only through Docling) → correctly 415'd by the #975 pre-flight, but the docs (#1121) currently imply they work.
- **Crawl4AI is bundled** (#432): every image carries Chromium (~300 MB) even for users who never touch it.

Baking Docling into the default image is far heavier than Crawl4AI (torch/transformers) and wrong for a self-hostable, privacy-first product. A second "full" image tag was considered and rejected for now (release/registry complexity). Env-gated startup install keeps a single lean image while making both runtimes available on demand.

## Decision

**Single lean image + env-gated startup install, with download caches persisted on the data volume (not the venv).**

### Env vars (two booleans)
- `OPEN_NOTEBOOK_ENABLE_DOCLING=true` → install `content-core[docling]`; unlocks the `docling` document engine, the OCR toggle, and image-source support.
- `OPEN_NOTEBOOK_ENABLE_CRAWL4AI=true` → install `content-core[crawl4ai]` + the Chromium browser; unlocks **local** Crawl4AI.
- **Crawl4AI remote nuance:** if `CRAWL4AI_API_URL` is set, the `crawl4ai` engine works against a remote server with **no local install**. So "crawl4ai available" = (local package installed) **OR** (`CRAWL4AI_API_URL` set). Only local mode triggers the install.

### Startup / install mechanics
- New **entrypoint** wrapping supervisord (currently `CMD` runs supervisord directly). It runs **before** `api`/`worker` so they only start once extras are ready (blocking, no supervisord-ordering hacks).
- **Idempotent**: a marker on the volume records the installed extras + a version stamp (Python version + pinned `content-core` version). If the stamp matches, skip; if it mismatches (image upgraded), reinstall from cache. Self-healing across upgrades.
- **Degrade, don't die** (confirmed): if an install fails (e.g. no network), log **loudly** and start anyway — the app runs, the extra is reported unavailable. A PyPI hiccup must not take the whole app down.
- **Loud, honest logs**: first boot with a flag prints e.g. `Installing Docling (first start; this downloads ~2 GB and can take several minutes)…`.
- Both `api` and `worker` share `/app/.venv`, so a single install serves both.

### Persistence (the important part — caches on volume, NOT the venv)
Persist only the **download caches** on `/app/data` and reinstall into the **image-layer venv** from cache each boot:
- `UV_CACHE_DIR=/app/data/.cache/uv` → wheels cached; boot reinstall is a local copy, no re-download.
- `PLAYWRIGHT_BROWSERS_PATH=/app/data/.cache/playwright` → Chromium persists (identical path at install and run — the Playwright gotcha).
- `HF_HOME=/app/data/.cache/huggingface` → Docling downloads its models (DocLayNet/TableFormer) on **first use**, not at install; this persists them.

Rationale: persisting the venv itself (ComfyUI/A1111 style) couples to the image's Python version and breaks on upgrades. Reinstalling into the fresh image venv from volume-cached wheels is immune to that **and** avoids re-downloads. Cost: ~10–30 s of wheel-copy per boot vs. GBs of download.

### Dockerfile changes
- **Un-bundle Crawl4AI (revert the #432 bake):** `pyproject.toml` `content-core[crawl4ai]` → `content-core`; remove the `playwright install chromium` step and the image-layer `PLAYWRIGHT_BROWSERS_PATH=/app/pw-browsers`.
- Add the entrypoint to `runtime-base` so both the `single` and `runtime` variants inherit it.
- Keep `uv` available and the venv writable for runtime installs. Revisit `UV_NO_SYNC=1`.
- `playwright install --with-deps chromium` needs apt (root) at runtime for system libs — acceptable on first boot; document it.

### Backend: capability probe
- Expose **real availability** (not the flag): `GET /api/capabilities` (or fold into an existing config/settings endpoint the UI already fetches, à la `GET /api/providers`).
- Checks: `content_core…docling.DOCLING_AVAILABLE` (import), a `crawl4ai` import; `crawl4ai` effective = local import **OR** `CRAWL4AI_API_URL` set.
- `graphs/source.py` already falls back gracefully (crawl4ai→None continues the auto chain; docling absent→pdfplumber/etc.), so nothing crashes without the extras — the probe is for UI honesty, not crash-avoidance.

### Frontend: UI gating (confirmed: disabled + hint)
- Settings → Content Processing: render the `docling` and `crawl4ai` engine options and the **Enable OCR** toggle, but `disabled` when the capability probe reports them unavailable, with a hint naming the env var (e.g. *"Requires `OPEN_NOTEBOOK_ENABLE_DOCLING=true` — installs on first startup"*).
- Source of truth = the capability probe (actual availability), **not** the env var — so while first-boot install is still running, the option correctly shows unavailable.
- A short note in the card explaining these are optional, heavier engines.
- This fixes the current dead-UI behavior (selectable option that silently falls back).

### #975 interaction
`check_file_support` already reflects Docling availability (same routing), so unsupported-image → 415 stays correct automatically. Nice-to-have: when an image is rejected and Docling is off, the message can hint "enable Docling for image OCR".

### Reproducibility
Pin the runtime install to the locked `content-core` version (`content-core[docling]==<locked>`), so the extra only adds compatible deps and doesn't drift from the base install.

## Amends prior work
- **#432** — un-bundles Crawl4AI from the default image (feature stays; "bundled by default" is reverted).
- **#1104** — the OCR toggle gains availability gating (stops being a silent no-op).
- **#1121** — docs updated to describe Docling/Crawl4AI/OCR/image-sources as **opt-in** with the enable vars and first-boot caveat (see below).

## Docs
- New/updated: `docs/5-CONFIGURATION/environment-reference.md` — `OPEN_NOTEBOOK_ENABLE_DOCLING`, `OPEN_NOTEBOOK_ENABLE_CRAWL4AI`; note the first-boot download + volume caching.
- `docs/3-USER-GUIDE/content-processing-engines.md` — an "Optional engines (Docling, Crawl4AI local): how to enable" section.
- `docs/3-USER-GUIDE/adding-sources.md` — image/OCR support noted as requiring Docling enabled.
- **New ADR (ADR-007)**: "Heavy extraction runtimes are opt-in, installed at startup." Records this structural decision and its rationale (image size, self-hostable posture, caches-on-volume). This also resolves the cubic finding on #1121 (edit the immutable ADR-002 → write a new superseding/complementary record instead).

## Offline / air-gapped (#374)
Startup install needs network on first boot. Since it's opt-in, air-gapped operators simply don't enable it. Document this; a future build-arg to pre-bake the extras for air-gapped builds is out of scope here.

## Testing
- Unit: the capability probe; the UI gating (disabled states from probe).
- Integration/smoke (`scripts/release-test`): boot with `OPEN_NOTEBOOK_ENABLE_DOCLING=true` / `_CRAWL4AI=true`, assert the engines become available and a second boot reuses the cache (no re-download); boot without flags → lean, options disabled.

## Acceptance criteria
- Default image ships **without** Docling or Chromium; base extraction (text PDFs, Office, EPUB, URLs via bs4/jina/firecrawl, audio, YouTube) works unchanged.
- Setting `OPEN_NOTEBOOK_ENABLE_DOCLING=true` and/or `OPEN_NOTEBOOK_ENABLE_CRAWL4AI=true` installs the runtime on first boot, caches on the volume, and a second boot does **not** re-download.
- With Docling enabled: `docling` engine + OCR toggle work; scanned PDFs and image sources extract.
- With Crawl4AI enabled (or `CRAWL4AI_API_URL` set): `crawl4ai` engine works.
- Install failure → app still starts, extra reported unavailable, loud log.
- UI shows unavailable engines/OCR disabled with an env hint; available ones enabled.
- Docs + ADR-007 landed.

## Out of scope / future
- A separate pre-baked "full" image tag.
- A build-arg to bake extras for air-gapped deployments.
- Exposing `docling_formulas` / `docling_vision` (tracked under #331).

## Related
- Amends #432, #1104; updates #1121. Part of the #939 (content-core 2.x) follow-up. Offline: #374.

lfnovo:
## What & why

Docling (layout parsing, OCR, image sources) and local Crawl4AI (JavaScript rendering via Chromium) are heavy runtimes that don't belong in every image. As shipped by the content-core 2.x epic (#939), Crawl4AI was **bundled** into the default image (~300 MB Chromium) while Docling was **never installed** — so the OCR toggle (#1104) was a silent no-op and image sources were unsupported. Baking Docling in would add a multi-GB ML stack (torch/transformers) to every image, wrong for a lean, self-hostable product.

This makes both **opt-in**, installed automatically on **first container startup**, keeping the default image lean. Design recorded in **ADR-007**.

## How

**Infra (un-bundle + startup install)**
- `pyproject.toml` drops the `[crawl4ai]` extra; `uv lock` regenerated (crawl4ai/docling/playwright out of the base image).
- `Dockerfile`: removes the build-time `playwright install`; adds an `ENTRYPOINT` and points `UV_CACHE_DIR` / `PLAYWRIGHT_BROWSERS_PATH` / `HF_HOME` at the `/app/data` volume.
- New `scripts/docker-entrypoint.sh`: env-gated, **blocking** install before supervisord, pinned to the base `content-core` version, **degrade-don't-die** with loud logs, idempotent across restarts. Downloads persist on the volume (second boot reinstalls from cache — no re-download); the venv stays in the image layer (immune to Python-version bumps).
- `docker-compose.yml`: documents the new opt-in env vars.

**Env gates**
- `OPEN_NOTEBOOK_ENABLE_DOCLING=true` → Docling engine + OCR + image sources.
- `OPEN_NOTEBOOK_ENABLE_CRAWL4AI=true` → local Crawl4AI. A remote server via `CRAWL4AI_API_URL` needs no local install.

**Backend probe**
- New `GET /api/capabilities` (`api/routers/capabilities.py`, `CapabilitiesResponse`) reporting *actual* availability: `docling_available`, `crawl4ai_available` (local **or** remote), `crawl4ai_remote_configured`.

**Frontend gating**
- `useCapabilities()` hook + `capabilities` API/type. `SettingsForm` disables the Docling engine, the Crawl4AI engine and the OCR toggle (with an env-var hint) until the runtime is available — source of truth is the probe, so it reads correctly even mid-install. Hint strings added across **all 14 locales**.

**Docs (folds in the held #1121, rewritten to opt-in)**
- `content-processing-engines.md` (new), `adding-sources.md`, `index.md`, `environment-reference.md`, `advanced.md`, and **ADR-007**.

## Tests
- Backend `tests/test_capabilities_api.py` (probe composition: local **or** remote ⇒ available). Full suite: 534 passed.
- Frontend `SettingsForm.test.tsx` (disabled states + hints flow from the probe; optimistic default while loading). Full suite: 122 passed; `lint`/`build` clean.

## Verification still needed (needs a real image build — CI/manual)
The Docker smoke path can't be exercised locally here:
1. Default image ships **without** Docling/Chromium; base extraction unchanged; Settings shows the three options disabled with hints.
2. Boot with `OPEN_NOTEBOOK_ENABLE_DOCLING=true` → loud install; `GET /api/capabilities` flips `docling_available: true`; scanned PDF + `.png` extract.
3. **Second boot** with the volume → no re-download.
4. `OPEN_NOTEBOOK_ENABLE_CRAWL4AI=true` (or `CRAWL4AI_API_URL`) → crawl4ai works / reported available with no local install.
5. Offline first boot with a flag → app still starts, loud log, option stays disabled.

Closes #1122. Closes #1105 (docs folded in). Supersedes #1121. Amends #432, #1104. Part of #939. Offline follow-up: #374.

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1123?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lfnovo:
Thanks — addressed all four review findings in 351ec65:

- **P2 · Crawl4AI probe could report "usable" without a browser** — local availability now requires the package **and** a Chromium browser on disk (`_crawl4ai_local_ready` → `_chromium_browser_present`, scanning `PLAYWRIGHT_BROWSERS_PATH`). Since the startup installer downloads the package and the browser in separate, degrade-don't-die steps, a package-present-but-browser-missing state now correctly reports **unavailable**. Added readiness tests (missing package / empty browser dir / browser present).
- **P2 · `except Exception` masking real errors** — narrowed both probes to `except (ImportError, AttributeError)` and log unexpected failures (`logger.opt(exception=True).warning(...)`) instead of silently falling back.
- **P3 · Brittle checkbox lookup** — the SettingsForm test now targets the OCR toggle by accessible name (`getByRole('checkbox', { name: 'settings.ocrEnabled' })`).
- **P3 · fr-FR / ca-ES copy** — fixed the French reflexive verb (`s'installe au premier démarrage`) and the Catalan feminine agreement (`en la primera arrencada`).

Backend + frontend suites green (`ruff`/`mypy` clean; locale parity + SettingsForm tests pass).

lfnovo:
## Review — opt-in Docling + Crawl4AI runtimes (#1122) 🔍 approve with 2 changes requested

Big, well-considered change. Reviewed manually + cubic CLI; verified each finding against the code and the installed `content-core==2.0.4`. The strategy (un-bundle both heavy runtimes, install on first boot, probe actual availability, gate the UI) is sound and cleanly recorded in ADR-007. This also correctly resolves the #1121 docs gap — the docs now frame Docling/Crawl4AI as opt-in, the OCR toggle as inert-until-Docling, and image sources as rejected when Docling is off.

### Verified solid
- **Entrypoint** (`scripts/docker-entrypoint.sh`): POSIX `/bin/sh`, `has_module` probes the real venv each boot, extras pinned to the installed content-core version, degrade-don't-die, idempotent, caches on `/app/data` (volume) while the Playwright system-deps stamp lives on the container FS (correct — system libs reset per container, the browser persists). `ENTRYPOINT` set in `runtime-base` so both image variants inherit it; `exec "$@"` hands off to supervisord.
- **Probe** (`api/routers/capabilities.py`): `_docling_available()` reads content-core's authoritative `DOCLING_AVAILABLE` routing gate (not a naive spec check), and `crawl4ai_available = local OR remote` is right. Boot ordering is correct — the blocking install finishes before the API imports `content_core`, so no stale flag.
- **Frontend**: `useCapabilities` caches like providers (`staleTime: Infinity`); gating disables the Docling/Crawl4AI engine options and the OCR toggle with per-locale hints (14 locales). Backend `test_capabilities_api.py` green (local-or-remote composition).

### 🔧 Change requested (2)

**1. (P1) Crawl4AI is reported available when only the package installed but Chromium failed.** The entrypoint installs `content-core[crawl4ai]` and *separately* runs `playwright install --with-deps chromium`, and by design continues on a Chromium failure. But the probe only checks `find_spec("crawl4ai")`, so in that partial state `crawl4ai_available: true` and the UI enables the engine — while local rendering fails (`AsyncWebCrawler` can't launch → `extract_url_crawl4ai` returns `None` → empty extraction on explicit selection). This contradicts ADR-007's own promise that a failed install is "reported unavailable rather than crashing." Suggest the probe also verify the Chromium browser is present (e.g. check `PLAYWRIGHT_BROWSERS_PATH` for the chromium build, or a cheap `playwright` launch check) before reporting local Crawl4AI available. (Auto mode is unaffected — it falls back to bs4.)

**2. (P2) The frontend fails *open* on a probe error.** `capabilities?.docling_available ?? true` (and the Crawl4AI equivalent) treat *undefined* as available. That's the right optimistic default while **loading**, but it also triggers on a **failed** probe (404/500/network) — enabling engines the backend couldn't verify. Recommend distinguishing the query's `isError`/`isLoading`: keep the optimistic default only while pending, and fail closed (or show a pending/error hint) on error.

### Notes (no change required)
- **(P2) Docling `find_spec` fallback** — `_docling_available()` falls back to `find_spec("docling")` if importing `DOCLING_AVAILABLE` throws. Low-probability (content-core catches its own docling import internally), but the fallback is a weaker signal that could over-report a partial install. Fine to leave.
- **(P2) Runtime extras float their transitive deps** — `uv pip install content-core[docling]==<ver>` resolves transitive versions at boot, not from `uv.lock`, so identical images can drift over time. ADR-007 acknowledges this and pins the content-core version as partial mitigation; flagging only so it's a conscious, documented tradeoff.

### One risk to call out
CI builds the image but never runs the `ENTRYPOINT`, so the whole opt-in install path is unverified by automation — the 5 "verification still needed" scenarios remain manual. Worth a real image build+boot (default; `ENABLE_DOCLING`; second boot cache-hit; `ENABLE_CRAWL4AI`/remote; offline) before merge, since that's exactly where a shell/entrypoint bug would hide.

Architecture LGTM; the two requested changes tighten the "UI reflects reality" contract that is the whole point of the probe.

lfnovo:
Thanks for the thorough manual pass. Both change requests are handled:

- **Change 1 (P1) · Crawl4AI available without Chromium** — already addressed in 351ec65 (posted just before this review): local availability now requires the package **and** a Chromium browser on disk via `_crawl4ai_local_ready()` → `_chromium_browser_present()` (scans `PLAYWRIGHT_BROWSERS_PATH` for the chromium build), exactly the check you suggested. Package-present-but-browser-missing now reports **unavailable**, restoring ADR-007's promise. Readiness tests cover missing-package / empty-browser-dir / browser-present.
- **Change 2 (P2) · Frontend failed *open* on probe error** — fixed in a72756e. The optimistic default now applies only while the query is **pending**; on `isError` it **fails closed** (`capabilities?.docling_available ?? !capabilitiesError`), so a failed probe no longer advertises unverified engines. Added a regression test asserting hints show + OCR disabled on probe error.

On the two "no change" notes and the CI/entrypoint risk — agreed on all: the `find_spec` docling fallback stays (weak but harmless), the runtime-extras version-float is a documented ADR-007 tradeoff, and the 5 entrypoint smoke scenarios remain manual (a real image build+boot before merge is the right gate — CI builds the image but never runs the ENTRYPOINT).

lfnovo:
Fixed in 259905b. Valid catch — the dev branch (no `PLAYWRIGHT_BROWSERS_PATH`) trusted the package alone, but resolving Playwright's default path never downloads Chromium, so a package-only dev install was over-reported as usable. `_chromium_browser_present()` now resolves the per-user default cache (`ms-playwright`, platform-aware) when the env var is unset and requires a chromium build there too, failing closed otherwise. Docker is unaffected (it always sets `PLAYWRIGHT_BROWSERS_PATH`). Added dev-cache present/absent tests (9 capability tests total).

lfnovo:
## ✅ Docker smoke — all 5 scenarios pass

Built the `single` target locally from `259905b` and exercised the entrypoint install path (which CI never runs). Used **local Crawl4AI** as the real install proxy (its entrypoint path is a superset of Docling's — `uv pip install` + `playwright install --with-deps` — at a fraction of Docling's multi-GB torch download); Docling's own install/degrade/gating was validated via the offline + probe scenarios.

**Image is lean** — venv **805 MB**, and `torch`/`transformers`/`crawl4ai`/`docling`/`playwright` are all **absent** from the base (un-bundle confirmed). Entrypoint present + executable; `UV_CACHE_DIR`/`PLAYWRIGHT_BROWSERS_PATH`/`HF_HOME` point at `/app/data`.

| # | Scenario | Result |
|---|----------|--------|
| A | Default boot, no flags | `/api/capabilities` → all `false`; entrypoint logs only `Starting Open Notebook`; crawl4ai/docling absent from venv |
| B | `ENABLE_CRAWL4AI=true` first boot | Blocking install (~55s), loud logs `Installing… → package installed → Chromium + system libraries ready`; probe flips `crawl4ai_available: true`; chromium build + 549 MB uv cache land on the volume |
| C1 | Restart same container | Idempotent skip: `Crawl4AI package already installed` → `Chromium browser ready` (no `--with-deps`, no download) |
| C2 | Fresh container + same volume | Boot **~12s** (vs 55s), **0** "Downloading" log lines, chromium dir mtime **unchanged** → wheels from uv cache, browser from volume, no re-download |
| D | `ENABLE_CRAWL4AI=true` + `CRAWL4AI_API_URL` | `using the remote Crawl4AI service (no local install)`; probe `crawl4ai_available: true` + `crawl4ai_remote_configured: true`; package **not** installed |
| E | `ENABLE_DOCLING=true`, `--network none` | Install fails → **loud** `WARNING: Docling install FAILED. Booting without it…` → app still **healthy**; probe `docling_available: false` (degrade-don't-die) |

Two asides (neither introduced by this PR, both out of scope):
- The local image measured 8.79 GB only because `COPY . /app` pulled in leftover `.claude/worktrees/**` (each with a full `.venv` + `node_modules`) — `.dockerignore` covers root `.venv`/`frontend/node_modules` but not `.claude/`. CI builds from a clean checkout, so published images are unaffected. Optional follow-up: add `.claude` to `.dockerignore`.
- `single`-container boot needs `SURREAL_USER`/`SURREAL_PASSWORD` to match the bundled SurrealDB (pre-existing).

The opt-in machinery — blocking install, loud logs, volume cache reuse, remote-no-install, and degrade-don't-die — all behave as designed.

PR #1130, opened 2026-07-14
feat: parallel chunk processing for large-document transformations (#990)

### LINKED ISSUE #990: [Feature]: parallel chunk processing for large-document transformations
## Problem

Transformations run the whole source through the model in one call. For documents larger than the model's context window, small/cheaper models fail outright — today you need an expensive large-context model to transform big sources.

## Proposed feature

Let transformations handle arbitrarily large documents automatically:
1. Optimistically try the full content.
2. On a context-limit error, split into token-sized chunks and process them in parallel (LangGraph `Send`).
3. Synthesize the partial results into one output (hierarchically, so combining many results doesn't overflow the window either).

Zero configuration — it only kicks in when the full content doesn't fit; small documents keep the current single-pass behavior.

## Prototype

There's a working, tested implementation in **#529** (thanks @kevincolten). It's in review; formalizing the feature here so the PR has an approved issue to reference.

## Design / review notes (from the #529 review)

- Must **not** regress the non-chunked path — keep the current `max_tokens` output budget (8192) for the full-content attempt.
- Avoid module-level `asyncio` primitives (event-loop binding footgun with the worker).
- Context-limit detection currently parses provider error strings — document supported formats + fallback behavior.
- Keep transformation semantics intact for non-summary transformations (chunk meta-hints shouldn't leak into output).
- Mind the interaction with worker concurrency (`OPEN_NOTEBOOK_WORKER_MAX_TASKS`, #893).

## Acceptance criteria

- A transformation on a document larger than the model's context completes successfully on a small-context model.
- Small-document transformations are unchanged (output and token budget).
- No new module-level event-loop-bound state.

dyzur:
Implements automatic parallel chunk processing for large-document transformations.

When a document exceeds the model's context window, the graph automatically:
1. Falls back from full-content attempt on context-limit error
2. Splits the content into token-bounded chunks (paragraph -> sentence -> word hierarchical splitting)
3. Processes chunks in parallel via LangGraph Send
4. Hierarchically synthesizes results into a single coherent output

For small documents, behavior is identical to the original single-node graph (full content, 8192 output token budget).

Closes #990

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1130?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lfnovo (changes requested):
Full review run on top of the cubic pass. The five P1s cubic raised are all genuinely addressed in `f617006` — verified against the current diff, not re-litigating them.

One new **P1** blocks merge, plus some P2s worth a look.

### P1 (blocker)

**`max_tokens` mismatch defeats the feature on sub-~82k-context models** — `open_notebook/graphs/transformation.py:137-140` vs `:285` (and merges at `:395`, `:490`).

Chunks are sized reserving `output_budget = min(FULL_CONTENT_MAX_TOKENS, int(context_limit * 0.10))`, so for any effective context below ~82k, `output_budget < 8192` and `max_chunk_tokens = context - system - output_budget`. But the actual chunk/merge LLM calls then request `max_tokens=FULL_CONTENT_MAX_TOKENS` (8192, hard-coded). That makes `chunk_input + 8192 + system > context_limit`, so providers that count requested `max_tokens` against the window reject the call — the chunk re-fails with the *same* context-limit error, `process_chunk` raises, and the whole transformation aborts.

This breaks the feature precisely for the small/mid-context models it exists to serve (e.g. a 32k model on a 40k-token doc). Fix: have the chunk and merge calls request `max_tokens=output_budget` instead of the constant.

### P2 (worth addressing, discussable if here or follow-up)

- **`section_hint` is dead code** — `transformation.py:247-252`. `_build_system_prompt` injects `section_context`, but `prompts/transformation/execute.jinja` only renders `{{ instructions }}`, so the "section X of Y" hint never reaches the model. Chunks are processed with no partial-document awareness.
- **Progress-guard has no real test coverage** — `transformation.py:387`. `test_progress_guard_fallback_on_single_item_groups` uses ~2000-token texts against a ~6873-token grouping threshold, so groups always hold multiple items and the fallback branch (the point of the anti-hang guard) never executes.
- **Unbounded fan-out concurrency** — `transformation.py:216`. One `Send` per chunk with no `max_concurrency`; N scales with document length → rate-limit storms, compounded across concurrent worker jobs. Issue #990 explicitly asked to mind `OPEN_NOTEBOOK_WORKER_MAX_...`.
- **One chunk failure discards all sibling work** — `transformation.py:266`. `process_chunk` re-raises on any exception; with parallel `Send`, a single transient error (e.g. a rate limit) aborts the entire transformation with no retry or partial-result path.
- **Concatenation fallback leaks scaffolding into the stored insight** — `transformation.py:~430-450`. When no adjacent pair fits the window, texts are joined with literal `Part N:` / `---` markers and returned as the final output with no LLM merge — meta-structure leaks into the persisted insight for non-summary transformations (the one place issue #990's "meta-hints shouldn't leak" requirement is actually violated).

Great work overall — the chunking hierarchy and hierarchical synthesis are solid, and the cubic fixes landed cleanly. Just the P1 to sort before merge.

lfnovo (changes requested):
Re-reviewed against the current branch. The `max_tokens` blocker from my earlier pass is genuinely fixed (chunk/merge calls now derive from `output_budget` at L294/425/532), and `section_hint` reaches the model and the scaffolding no longer leaks into the stored insight. But one P0 remains — and the tests that claim to cover it don't.

### P0 — the oversized-result re-chunk branch is dead code
`open_notebook/graphs/transformation.py:485`. The `raise ConfigurationError(...)` inside the `if 2 * re_chunk_budget + overhead > context_limit:` block is unconditional, so everything after it (the `logger.warning`, the `re_chunked` build loop, `texts = re_chunked`, and `continue` at L492–515) is unreachable. Consequences:

- When the guard is **false** (the normal case), control skips the whole `if` and falls straight to the forced pairwise merge at L516+, which sends `Part 1 + Part 2` (= `min_combined` tokens) to the provider. That branch is only reached *because* grouping found no adjacent pair fits `context_limit - output_budget - 500`, i.e. `min_combined` is large — so the oversized pair is shipped with no guard and the context-limit rejection recurs on ordinary sub-~82k windows. This is exactly the failure the feature exists to prevent.
- The re-chunk remediation the "re-chunk oversized results" / "prevent hang" commits describe never actually executes.

Fix: make the `raise` conditional on genuine infeasibility and let the re-chunk block run (and `continue`) when re-chunking *is* feasible — i.e. the re-chunk loop needs to sit on the reachable path, not after the `raise`.

### Tests don't cover the re-chunk path
`tests/test_graphs.py` — `test_stalled_synthesis_rechunk_fallback` and `test_progress_guard_fallback_on_single_item_groups` are tautological: with the default `context_limit=8192` and their inputs, the L484 guard is always false, so the re-chunk path (dead code anyway) never runs; both simply loop over a mock that returns a fixed string and would pass whether or not re-chunking works. Please add a test that drives the re-chunk/guard path with a **small `context_limit`** and asserts re-chunking actually occurred (and that no oversized pair is forwarded to the merge). `process_chunk` and the full `try_full_content → context-error → chunk → synthesize` flow are also currently untested.

### P2 (discussable — here or follow-up)
- **Concurrency ignores `OPEN_NOTEBOOK_WORKER_MAX_TASKS`.** `MAX_CONCURRENT_CHUNKS = 10` is hard-coded (L52) and the `Send` fan-out has no `max_concurrency`; #990 explicitly asked to honor the worker cap. Also, the module-level `asyncio.Semaphore` is bound to the import-time event loop.
- **Still fail-all on a single chunk error.** `synthesize_results` now aggregates a classified error instead of aborting immediately (an improvement), but no partial insight is ever saved and there's no per-chunk retry. Whether a partial-result path is in scope is your call.

The chunking hierarchy and hierarchical synthesis are solid — it's the P0 (dead re-chunk branch) plus a real test for it that block merge. The two P2s are design calls I'm happy to defer to a follow-up if you'd rather land the core first.

Issue #87, opened 2025-06-30
Request: select notebook in ask and search


rubujubi:
Currently, ask and search query the entire knowledge base (all notes created and all documents uploaded).

Suggestion: add a drop down menu to enable selecting notebooks that we want to include in ask and search.

lfnovo:
That's a good suggestion. Thanks. Will track and find the best momento to implement.

saketbyte:
Hey there! Is there any work going around this? Also wanted to know if we could do otherwise as well, instead of attaching notebook level scope in ask and search, can we introduce embedding-level vector  search in notebook itself? At present, from what I see, is that we are sending "full content" or "full text" means entire file content with each query. And that takes up time and tokens for each query asked. Can we introduce vector search at the chat/execute or context end point? 

I also noticed that when we are generating source embeddings here - https://github.com/lfnovo/open-notebook/blob/main/commands/embedding_commands.py#L319

Can we attach table name as source id instead of "source_embedding", and then perform a vector search on the tables/notebooks associated with a specific notebook only, this allows vector search on embeddings per query and fetches relevant data only?

lfnovo:
## Scope for implementation

Add an **optional notebook filter** to the ask and search features.

**In scope:**
- Add optional `notebook_id` (or list of notebook IDs) to `SearchRequest` and `AskRequest` (`api/models.py`).
- Thread it through `search_knowledge_base` (`api/routers/search.py`), the ask graph (`open_notebook/graphs/ask.py`), and the domain functions `vector_search`/`text_search` (`open_notebook/domain/notebook.py`).
- Add a `WHERE` clause filtering by the source↔notebook / note↔notebook relationship in the SurrealQL functions `fn::text_search` / `fn::vector_search` (`open_notebook/database/migrations/`).
- Add a notebook selector (multi-select, defaults to all) in the search/ask UI (`frontend/src/app/(dashboard)/search/page.tsx`).

**Acceptance criteria:**
- When one or more notebooks are selected, ask/search only return results from those notebooks.
- When none are selected, behavior is unchanged (search the entire knowledge base).

**Out of scope:** the per-notebook embedding-table redesign proposed in the comments — if desired, that should be filed as a separate issue.

lfnovo:
[...]

PR #1142, opened 2026-07-14
feat: add notebook filter to ask and search (#87)

dyzur:
Implements #87 — filter ask and search results by notebook.

### Backend
- **New migration 23** — redefines `fn::text_search` and `fn::vector_search` with an optional `$notebook_ids` parameter. Pre-computes valid source/note record IDs from the `reference` (source↔notebook) and `artifact` (note↔notebook) relationship tables, then filters with the `INSIDE` operator.
- **API models** — `notebook_ids: Optional[List[str]]` added to both `SearchRequest` and `AskRequest`
- **Domain functions** — `text_search()` and `vector_search()` accept and pass through `notebook_ids`
- **Router** — threads notebook_ids through search endpoint and both ask endpoints (streaming + simple)
- **Ask graph** — `provide_answer` reads `notebook_ids` from `configurable` and passes to `vector_search`

### Frontend
- **NotebookMultiSelect** — new Command+Popover+Checkbox multi-select component with search filtering and badge display
- **Search page** — notebook filter UI added to both Ask and Search tabs
- **use-ask hook** — updated to accept and pass `notebook_ids`
- **i18n** — new keys added to all 14 locales

## Behavior
- No notebook selected → search all (fully backward compatible)
- One or more notebooks selected → results scoped to those notebooks
- Empty array / nonexistent IDs → no results

## Testing
Verified end-to-end with 4 test scenarios against live SurrealDB:
- Unfiltered search returns results from all notebooks
- Filtered search returns only results from the specified notebook
- Cross-notebook isolation works (no spillage)
- Nonexistent notebook filter returns empty

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1142?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lfnovo (changes requested):
Thanks for this — the feature is well-structured and genuinely delivers #87: a notebook multi-select wired into **both** Ask and Search, threaded end-to-end to a notebook-aware SurrealQL filter. I verified the parts that are easy to get wrong and they're correct: `notebook_ids` is passed as a bound `$notebook_ids` param (each id via `ensure_record_id()` → `RecordID`), so there's no injection surface; all DB access stays `await`-ed; `None`/empty/nonexistent-id handling short-circuits correctly; the `reference` (source→notebook) / `artifact` (note→notebook) relation directions are right; and the 4 i18n keys landed in all 14 locales. `async_migrate.py` wires migration 23 into both up and down lists.

Two bugs in the rewritten search functions block merge, plus a coverage gap.

### P0 — `vector_search` drops migration 9's dimension/none guard
`migrations/23.surrealql:813,829,846` (and identically in `23_down.surrealql`). Migration 9 exists specifically to add this guard to all three sub-searches:
```
WHERE embedding != none and array::len(embedding)=array::len($query) AND
      vector::similarity::cosine(embedding, $query) >= $min_similarity
```
Migration 23 reverts to the pre-9 body (`WHERE vector::similarity::cosine(...) >= $min_similarity` only). `vector::similarity::cosine` requires equal-length vectors; on records whose stored embedding is `none` or a different dimension — the normal state after switching embedding model/provider, exactly what the guard was added for — this errors and fails the entire vector search. It's hit by `POST /search` (vector mode) and every `ask` request (`graphs/ask.py::provide_answer` → `vector_search`). Please carry migration 9's guard prefix into all three WHERE clauses of both up and down.

### P1 — `text_search` `$source_full_search` selects `source.id`/`source.title` FROM source
`migrations/23.surrealql:742` (and `23_down.surrealql:900`). This block is `FROM source`, where there is no `source` field, so `source.id`/`source.title` resolve to NONE and the rows are dropped by the final `where id is not None` — breaking full-text body search for sources (the only content matcher when no embedding model is configured). Migration 4 and the sibling `$source_title_search` in the same function use bare `id`/`title`. Note the adjacent `$source_embedding` block (`:733`, `FROM source_embedding`) correctly uses `source.id` — the bug looks like that pattern got copied into the `FROM source` block. Revert to `SELECT id, title, ... id as parent_id FROM source`.

### P1 — down migration isn't a faithful revert
`23_down.surrealql:1-2` says "Reverts to migration 9 versions", but it inherits both bugs above, so a rollback lands on a state buggier than what's live. Cleanest fix: make the down `text_search` = migration 4 verbatim and `vector_search` = migration 9 verbatim, then apply the `$notebook_ids` additions only to the up.

### P1 — no test exercises the actual filtering
`tests/test_search_api.py:68` only changed a pass-through assertion. Please add a test that verifies scoping really filters: source-in-notebook included, source-in-another-notebook excluded, nonexistent id → empty, `None` → unfiltered. That's the exact `INSIDE` logic carrying the two bugs above.

### P2 — description contradicts empty-array behavior
`api/models.py:9-15` maps `[]` → `None` (empty filter = all notebooks), but the PR body says "empty array → no results". The behavior is the sensible one; the description just needs correcting.

One cross-PR note: the stray `notebook_ids` fields that #1145 accidentally added to `api/models.py` are a fragment of this feature — this PR is their legitimate home, and whichever merges first will conflict the other. I've asked #1145 to drop them.

The cubic flag on pt-BR:517 is a false positive — that locale block is correct.

lfnovo (changes requested):
Following up on my 2026-07-18 review — the four blockers there are still open (no commits on the branch since 2026-07-14), and `main` has since moved in a way that adds one more. Summary of what's needed before this can merge:

**Still open from the 2026-07-18 review** (verified again against the current branch):
- **[P0]** `vector_search` in `migrations/23.surrealql` drops migration 9's guard. The three vector sub-searches are back to `WHERE vector::similarity::cosine(embedding, $query) >= $min_similarity` (lines 121/137/154), missing `embedding != none and array::len(embedding)=array::len($query) AND …`. This errors on records whose stored embedding is `none` or a different dimension — the normal state after switching embedding model/provider — and takes down both `POST /search` (vector mode) and every `ask` request. Carry migration 9's guard prefix into all three WHERE clauses (up and down).
- **[P1]** `text_search` `$source_full_search` block uses `source.id as id, source.title as title` under `FROM source` (line ~50), where there is no `source` field, so those rows get dropped by the final `where id is not None`. Use bare `id, title` here (as the sibling `$source_title_search` blocks do at lines 68/77). The adjacent `$source_embedding` block is `FROM source_embedding`, where `source.id` is correct — looks like that pattern got copied one block too far.
- **[P1]** `23_down.surrealql` inherits both bugs, so a rollback lands on a state buggier than what's live. Make the down `text_search` = migration 4 verbatim and `vector_search` = migration 9 verbatim, then apply only the `$notebook_ids` additions to the up.
- **[P1]** No test exercises the actual filtering. Please add a test that asserts scoping really filters: source-in-notebook included, source-in-another-notebook excluded, nonexistent id → empty, `None` → unfiltered.
- **[P2]** `api/models.py` maps `[]` → `None` (empty filter = all notebooks), but the PR body says "empty array → no results". The behavior is the right one; just correct the description.

**New — migration number collision:** `main` now ships migration **23** (`Expose Docling formula & vision enrichment toggles`, merged via #1187 on 2026-07-19). This PR also adds `23.surrealql` / `23_down.surrealql`. Please rebase onto `main` and renumber this migration to **24** (and `24_down`), and wire `24` into both the up and down lists in `async_migrate.py`.

The feature itself is solid and genuinely delivers #87 — the multi-select is wired into both Ask and Search, the `$notebook_ids` binding is injection-safe, and the relation directions are right. It's the SurrealQL correctness in the rewritten search functions (P0 + the two P1s) plus the rebase/renumber that stand between here and merge.

(The cubic flag on `pt-BR/index.ts:517` is a false positive — that locale block is complete.)

Issue #238, opened 2025-10-31
[Feature]: use GPU to process source uploads


alpha754293:
### Feature Description

I just uploaded US government regulations to test out open-notebook's capability to be able to handle and process this type of a PDF. The PDF itself (for this part/section of US government regulations) is 1400 pages. Right now, it is using the CPU to process the file when it is uploaded.

Is there a way to use the GPU (RTX 3090) to process the files that have been uploaded?

I have ollama installed and up and running already.

Thanks.

### Why would this be helpful?

This would speed up the processing of files that have been uploaded to open-notebook.

caymann12:
I am using ollama locally and Nvidia GPU is being used when sources are uploaded to process.

I run ollama in a linux container and tested nvidia-smi is responding.
I made sure ollama has the LLMs loaded forever (normally is unloads in 4min by default)
I am running open notebook locally in a linux debian container.

I made sure ollama has LLM loaded (for example gemma3:latest for chat model) (also another embedding mode is loaded on GPU)
when open notebook makes APi call to ollama, the LLMs get to work.. 

I made sure ollama  api gets called by open notebook using the docker.env file (example: http://10.0.0.1:11434  in the docker.env)
I test the API is working by trying: http://10.0.0.1:11434/api/tags
I got a response of available LLMs etc under ollama.

Any changes to the docker.env file means, docker compose needs to be restarted to take effect.

lfnovo:
Closing — there are two separate stages here, and GPU only applies to one of them:

- **Embeddings / LLM calls** already use your GPU when you run Ollama (as @caymann12 confirmed — `nvidia-smi` shows the models working during processing).
- **File text extraction** (parsing the PDF) is handled by the content-core library and is **CPU-bound**. For plain-text PDFs, a GPU doesn't accelerate parsing. GPU acceleration only helps for **OCR / vision models** on scanned or image PDFs — and that path is tracked in #331 (Docling VLM integration).

So for a 1400-page text PDF, the slow part is CPU extraction, which a GPU can't speed up. If your regulations PDF is scanned/image-based, follow #331.

Closing as not separately actionable — reopen if you have a specific extraction-stage GPU use case not covered by #331.

PR #1161, opened 2026-07-18
feat: add PayPerQ (PPQ) provider

mattius459:
## Summary

Adds **PayPerQ (PPQ)** to Settings → API Keys as a full multi-modal provider (LLM, embedding, STT, TTS), backed by the PPQ profile you merged in [esperanto#238](https://github.com/lfnovo/esperanto/pull/238). PPQ is a pay-as-you-go gateway exposing hundreds of models from many labs through one OpenAI-compatible key.

Built entirely on the new provider registry — thanks for that refactor, it made this small:

- **`provider_registry.py`** — one `ProviderSpec` (all four modalities, `PPQ_API_KEY`, test model `auto`, catalog discovery `https://api.ppq.ai/v1/models`)
- **`api/models.py`** — `ppq` added to the `SupportedProvider` Literal (the one manual copy)
- **`model_discovery.py`** — bespoke `discover_ppq_models()`: the OpenAI-compatible catalog (chat → language) **plus** the documented embedding/STT/TTS models that `/v1/models` doesn't list (mirrors the Deepgram/ElevenLabs static-list pattern). Happy to drop the static append and rely on manual entry if you'd prefer the pure registry-only path.
- test sync lists + docs (provider index, setup guide, env reference, README, CHANGELOG)
- **No frontend change** — the Settings UI reads `GET /api/providers` at runtime.

## ⚠️ Draft — blocked on an esperanto release

This depends on an esperanto version that includes the PPQ profile (#238, merged to `main` but not yet in a published release — latest is v2.24.0). The `esperanto>=2.20.0,<3` pin allows it, but a fresh install today resolves to v2.24.0 (no PPQ), so PPQ won't create at runtime until esperanto ships.

Opening as a draft now so you can review and, if you like, **coordinate merging both together** — since you own both repos. Before marking ready:
- [ ] esperanto release cut with #238
- [ ] bump the `esperanto` pin floor to that release + refresh the lockfile
- [ ] mark ready

I'm happy to do the pin bump the moment esperanto is released, or you can.

## Verification

- Live discovery against api.ppq.ai returns the chat catalog as `language` + the 7 curated embedding/STT/TTS models, no misclassification.
- `55 passed` across `tests/test_credential_provider_validation.py` and `tests/test_credentials_api.py` (registry↔Literal↔discovery all in sync).

_Model IDs/endpoints match the esperanto #238 profile exactly (`base_url https://api.ppq.ai/v1`, embedding `openai/text-embedding-3-small`, STT `nova-3`, TTS `deepgram_aura_2`)._

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1161?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


mattius459:
Unblocked and **marked ready** ✅

esperanto **v2.25.1** is on PyPI with the PayPerQ profile (#238), so I've:
- bumped the pin to `esperanto>=2.25.1,<3` and refreshed `uv.lock` (2.22.0 → 2.25.1)
- verified PPQ end-to-end through the released esperanto: all four modalities register, live chat returns a completion, embeddings return a 1536-dim vector
- `55 passed` across the provider-validation and credentials tests

The dependency checklist in the description is now complete. Ready for review whenever you have a chance — thanks again!

mattius459:
Thanks @cubic-dev-ai — went through both:

**Issue 2 (discovery tests) — fixed, and it caught a real bug.** You were right that `tests/test_model_discovery.py` pinned the pre-PPQ provider sets. I:
- added `ppq` to both expected sets (`OPENAI_COMPAT_PROVIDERS` and `PROVIDER_DISCOVERY_FUNCTIONS`)
- added `TestPPQDiscovery` covering the static-modality augmentation — it asserts the chat catalog is classified as `language` (with an Amazon **Nova** LLM in the fixture to guard against a `"nova"`→speech-to-text misclassification) and that the documented embedding/STT/TTS models are appended
- while writing it, found a genuine bug: `discover_ppq_models` appended the static embedding/STT/TTS models **even with no `PPQ_API_KEY`**. It now returns `[]` when unconfigured, consistent with every other discoverer. Good catch prompting this.

Full suite is green now (71 passed across the discovery/validation/credentials tests).

**Issue 1 (`test_model="auto"`) — this one's a false positive; keeping `"auto"`.** `auto` isn't a sentinel here — it's PPQ's real, always-available smart-routing model. I verified the exact connection-test path end-to-end against the live API:
```
create_language("auto", provider="ppq").to_langchain().ainvoke(...) -> "OK"
```
Switching to `None` would actually **break** PPQ's connection test: `None` (dynamic/first-available) only works for providers with a bespoke tester branch (Ollama, openai_compatible). PPQ uses the generic Esperanto path, where `if not test_model:` returns `"No test model configured"`. So PPQ needs a concrete, always-valid model — and `auto` is the most robust choice (never 404s, never churns as the catalog changes). Happy to pin a specific chat model instead if preferred, but `auto` is intentional and verified.

mattius459:
Looks like you already landed PPQ in #1179 (and refined discovery in #1182 — nice use of `?type=all` to pull the non-chat modalities dynamically), plus the esperanto bump in #1176. So this PR is redundant — closing it. Thanks for getting PayPerQ in! 🙏

PR #1208, opened 2026-07-23
fix(frontend): silence noisy KaTeX warning on Unicode symbols in math spans

bitsandbots:
## Summary

AI-generated content often contains Unicode punctuation (en/em dashes, smart quotes) inside single-dollar math spans — e.g. a price range like `$5 – $10`. Models legitimately emit inline math as `$...$` per this project's own prompts (`prompts/*/system.jinja` "MATH FORMATTING" sections), so `remark-math` correctly reads that as inline math. KaTeX's default strict mode then logs a console warning for every character outside its symbol table (the en-dash, in this example).

The warning is harmless — KaTeX still renders a best-effort glyph — but noisy: I saw 66+ occurrences in a few days of normal self-hosted use.

## Fix

Pass KaTeX's own `strict` option as a function that ignores only the `unknownSymbol` error code, leaving every other strict check (e.g. deprecated commands) at its default. Extracted to one shared `KATEX_OPTIONS` constant (`frontend/src/lib/utils/katex-options.ts`) so the three markdown-rendering call sites (chat/insight/source rendering via `markdown-renderer.tsx`, the note editor live preview via `markdown-editor.tsx`, and the transformation playground) don't each duplicate the same override.

An earlier version of this fix disabled `remark-math`'s `singleDollarTextMath` option instead — that was wrong, since it breaks the intentional single-dollar inline math this project's prompts already rely on (caught by the existing KaTeX render test before it went further).

## Test plan

- [x] `npm run test -- --run` — 140/140 pass, including the existing test asserting single-dollar inline math (`$x^2 + y^2 = z^2$`) still renders via KaTeX
- [x] `npx tsc --noEmit` — clean
- [x] `npm run lint` — 0 errors (pre-existing unrelated warnings only)
- [x] `npm run build` — clean production build
- [x] Manually verified in a running instance: a note containing `$5 – $10` (the exact failure pattern) renders with zero console warnings, while `$x^2+y^2=z^2$` and `$$\int_0^1 x^2\,dx=\frac13$$` still render correctly via KaTeX/MathML

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1208?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/app/(dashboard)/transformations/components/TransformationPlayground.tsx
+++ b/frontend/src/app/(dashboard)/transformations/components/TransformationPlayground.tsx
@@ -16,6 +16,7 @@ import ReactMarkdown from 'react-markdown'
 import remarkGfm from 'remark-gfm'
 import remarkMath from 'remark-math'
 import rehypeKatex from 'rehype-katex'
+import { KATEX_OPTIONS } from '@/lib/utils/katex-options'
 
 interface TransformationPlaygroundProps {
   transformations: Transformation[] | undefined
@@ -128,7 +129,7 @@ export function TransformationPlayground({ transformations, selectedTransformati
                     <div className="prose prose-sm max-w-none dark:prose-invert">
                       <ReactMarkdown
                         remarkPlugins={[remarkGfm, remarkMath]}
-                        rehypePlugins={[rehypeKatex]}
+                        rehypePlugins={[[rehypeKatex, KATEX_OPTIONS]]}
                         components={{
                           table: ({ children }) => (
                             <div className="my-4 overflow-x-auto">
--- a/frontend/src/components/ui/markdown-editor.test.tsx
+++ b/frontend/src/components/ui/markdown-editor.test.tsx
@@ -1,4 +1,4 @@
-import { describe, it, expect } from 'vitest'
+import { describe, it, expect, vi, afterEach } from 'vitest'
 import { render } from '@testing-library/react'
 import MarkdownPreview from '@uiw/react-markdown-preview'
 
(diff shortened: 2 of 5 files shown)


bitsandbots:
Addressed both points in 6ce7c00:

1. Added `frontend/src/components/ui/markdown-editor.test.tsx` regression tests: one asserting the em/en-dash-in-price-range pattern that motivated this fix produces no `[unknownSymbol]` warning, and one asserting a different strict violation (`[commentAtEnd]`) is still warned on — proving the ignore is scoped to exactly one error code, not silently widened by a future edit.
2. Reworded the `KATEX_OPTIONS` comment in `katex-options.ts` to keep the metrics/layout caveat visible: KaTeX's parser falls back unrecognized codepoints to `mode: 'text'` rather than proper math-mode metrics, so suppressing the warning isn't purely cosmetic — an ignored symbol can render with approximated spacing. Kept scoped to stray punctuation, not a license to route arbitrary Unicode through math mode.

142/142 frontend tests pass (up from 140), `tsc --noEmit` clean, `eslint` clean.

lfnovo:
Reviewed and happy with this: scoping the `strict` override to the single `unknownSymbol` code changes console output only, and the shared `KATEX_OPTIONS` constant is the right shape. It now conflicts with #1294 (markdown editor follows the app theme), which landed yesterday and touched the same import lines in `markdown-editor.tsx` and its test. Could you rebase on main? Should be a trivial import-line resolution. I'll merge as soon as it's green.

PR #1209, opened 2026-07-23
Fix single-source chat missing full document content

mike-foucault:
## Problem

Chatting/asking questions about an individual source (not a notebook) silently failed to use the document's actual content when no transformation/insight had been generated for it — the LLM only ever saw the source's title, with no error or warning. Users would get generic "I have no document to analyze" style answers without understanding why.

## Root cause

`build_source_context()` (in `open_notebook/utils/context_builder.py`, used by the source-chat graph) requested `context_size="short"` from `Source.get_context()`. That mode never includes `full_text` — only `id`, `title`, `insights`. Without insights (i.e. no transformation run on the source), the context was reduced to just the title.

Switching to `context_size="long"` alone was not enough: it exposed a second, previously latent bug. The function's token-budget truncation loop dropped the **entire source** whenever its `full_text` alone exceeded `max_tokens` (very common for real documents, since the budget is 50k tokens) — leaving an empty context for exactly the documents most likely to have real content.

## Fix

- Use `context_size="long"` in `build_source_context` so `full_text` is always included, regardless of whether the source has insights.
- Fix the truncation loop to only drop insights (last-fetched first) for budget — the source itself is never dropped. The prompt formatter (`_format_source_context` in `open_notebook/graphs/source_chat.py`) already caps `full_text` at a fixed size before it reaches the LLM, so this doesn't risk oversized prompts.

## Tests

Added/updated tests in `tests/test_utils.py`:
- `test_source_and_insights_shape` now asserts `context_size="long"` is requested.
- New `test_keeps_source_when_it_alone_exceeds_budget`: a source whose `full_text` alone exceeds `max_tokens` is still returned instead of being dropped.
- Existing `test_truncates_insights_to_token_budget` and `test_missing_source_yields_empty_context` still pass unchanged in behavior.

```
tests/test_utils.py::TestBuildSourceContext::test_source_and_insights_shape PASSED
tests/test_utils.py::TestBuildSourceContext::test_truncates_insights_to_token_budget PASSED
tests/test_utils.py::TestBuildSourceContext::test_keeps_source_when_it_alone_exceeds_budget PASSED
tests/test_utils.py::TestBuildSourceContext::test_missing_source_yields_empty_context PASSED
4 passed
```

Also manually verified end-to-end locally: uploaded a document with no transformation, opened its chat, and confirmed the LLM's response now reflects the document's actual content instead of the "no document" fallback.

This is a small, self-contained bug fix (per [CONTRIBUTING.md](../blob/main/docs/7-DEVELOPMENT/contributing.md) — "small, obvious bug fixes" are exempt from requiring a prior Discussion/Issue).

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1209?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/open_notebook/utils/context_builder.py
+++ b/open_notebook/utils/context_builder.py
@@ -185,10 +185,18 @@ async def build_source_context(
         # Truncate to the token budget: drop insights from the end. The
         # source is never dropped, even if it alone exceeds the budget —
         # an empty context is worse than a source whose full_text gets
-        # capped later by the prompt formatter (see docstring).
+        # capped later by the prompt formatter (see docstring). If the
+        # source alone already exceeds the budget, trimming insights can't
+        # bring the total under it anyway, so keep them all instead of
+        # discarding useful context for nothing.
         total_tokens = sum(item_tokens)
-        if max_tokens:
-            while total_tokens > max_tokens and insights:
+        source_tokens = item_tokens[0] if sources else 0
+        if max_tokens is not None:
+            while (
+                total_tokens > max_tokens
+                and insights
+                and source_tokens <= max_tokens
+            ):
                 total_tokens -= item_tokens.pop()
                 insights.pop()
 
--- a/open_notebook/utils/context_builder.py
+++ b/open_notebook/utils/context_builder.py
@@ -166,7 +166,13 @@ async def build_source_context(
             source = None
 
         if source:
-            source_context = await source.get_context(context_size="long")
+            # Pass insights=[] so get_context() doesn't fetch and embed them
+            # itself — this function fetches and represents insights
+            # separately below. Fetching them here too would hit the DB
+            # twice and double-count their tokens toward the budget.
+            source_context = await source.get_context(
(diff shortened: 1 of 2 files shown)


lfnovo:
Thanks for this, you diagnosed the root cause correctly: `build_source_context` was requesting a short context and could drop the source itself during truncation. #1226 landed on main on 2026-09-02 fixing the same two issues and going a step further, with token-budgeted truncation of `full_text`, a reserved budget for insights, and a `source_text_status` signal the Source Chat graph now uses. Since this PR's change is a strict subset of that, I'm closing it in favor of #1226. Appreciate the work.

PR #1239, opened 2026-08-03
fix(podcasts): pre-flight speaker voices before generating a podcast

### LINKED ISSUE #1238: [Bug]: Podcast generation — four defects that make a Gemini/non-English run fail late with a misleading error

## Summary

On a Gemini/Vertex deployment with Hebrew (`he-IL`) sources, podcast generation burns several minutes and a full transcript's worth of tokens and then dies with an error that names the wrong cause. Four independent defects in this repo combine into that story; a fifth silently produces a podcast in the wrong language.

All of these were reproduced and fixed downstream before filing. Versions tested: `open_notebook` 1.14.0 (`sha256:e53f90d6153fcf4a64604d9a0c12cb0428a32cfb8dbcbb72e81ab12c013ee330`), `podcast-creator` 0.12.0, `esperanto` 2.25.1, `langchain-core` 1.4.7, outline/transcript model `google/gemini-3.6-flash`, TTS `google/gemini-3.1-flash-tts-preview`.

| # | Area | Issue |
| --- | --- | --- |
| 1 | `prompts/podcast/*.jinja` | The prompts hand the model a JSON skeleton it returns verbatim (`"speaker": "..."`), which then fails speaker-name validation |
| 2 | `prompts/podcast/*.jinja` | The app's templates shadow `podcast-creator`'s bundled ones and never reference `{{ language }}` — `episode_profile.language` silently does nothing |
| 3 | migration 7 seed data | Seeded speaker profiles ship OpenAI voice ids, unusable with a Gemini `voice_model`; the failure only surfaces after the whole transcript is generated |
| 4 | `commands/podcast_commands.py` | The single error hint covers one case and misleads on the two most common real failures |

Related upstream fixes that cannot be applied in this repo (filed/filable separately): `podcast-creator` classifies `OutputParserException` as non-retryable (`retry.py`) and renders `format_instructions` from the unvalidated `Transcript` model, and `esperanto`'s Google TTS provider collapses a bad-voice 404 into `Google API error: Requested entity was not found.`

1. Configured an episode profile with 6 segments, a 3-speaker speaker profile (`business_panel`), Gemini models.
2. Generated a podcast from Hebrew sources.

### How did it break?

Segment 2 of 6 aborted with:

```
Error executing command open_notebook.generate_podcast: Failed to parse
ValidatedTranscript from completion {"transcript": [{"speaker": "...", "dialogue": "..."}]}.
Got: 1 validation error for ValidatedTranscript
transcript.0.speaker
  Value error, Invalid speaker name '...'. Must be one of: Marcus Thompson, Elena Vasquez, Johny Bing
```

Segment 1 had already succeeded and was discarded with the outline — the episode row is only written after `create_podcast()` returns.

That completion is not a hallucination. It is the prompt's own example, returned verbatim. Three things make copying likely:

- `prompts/podcast/transcript.jinja:77-87` shows a fill-in skeleton with `"speaker": "[Actual Speaker Name]"` and a bare `...` continuation; `:122-124` shows a second copyable example `{"transcript": [...]}`. `prompts/podcast/outline.jinja:42-63` and `:79-81` do the same with `"[Segment Name]"` / `{"segments": [...]}`. A `...` inside a JSON example is indistinguishable from `...` as a *value*.
- The prompt contradicts itself about code fences: `transcript.jinja:77` opens a ` ```json ` block, `transcript.jinja:121` says "Do NOT wrap the JSON in ```json code blocks". Contradictions push the model toward pattern matching over instruction following.
- The schema the model receives carries no speaker enum: `{{ format_instructions }}` is rendered from `podcast-creator`'s **unvalidated** `Transcript` model, while parsing at runtime uses `ValidatedTranscript` with its speaker whitelist. The allowed names reach the model only as prose — the contract shown is not the contract enforced. (That last part is a `podcast-creator` change.)

### Suggested fix

In both app templates: render the example from the real speaker names so a verbatim copy is *valid* (guarding `speaker_names[1]` against being out of range, since Jinja's default undefined would render an empty string and put an invalid example in the prompt); ban placeholders explicitly (`...`, `…`, `[like this]`, `TODO`, empty strings, truncated lists); drop the ` ```json ` fence and the second `{"transcript": [...]}` / `{"segments": [...]}` example so the "no code fences" instruction stops contradicting the example above it.

Set `language: "he-IL"` on the episode profile, with Hebrew sources.

### How did it break?

The outline and all segment titles came back in **English**.

`podcast-creator` resolves templates in this order (`podcast_creator/config.py:233-262`): inline `templates` config → `prompts_dir` config → **`Path.cwd()/prompts/podcast/<name>.jinja`** → package resources. This app only calls `configure("speakers_config", …)` / `configure("episode_config", …)` (`commands/podcast_commands.py:285-286`), and the container's working directory is `/app`, so step 3 wins: `prompts/podcast/{outline,transcript}.jinja` shadow the library's copies.

The app copies never reference `{{ language }}`. The bundled ones do (`podcast_creator/resources/prompts/podcast/transcript.jinja:30-34`, `outline.jinja:28-30`). The value is plumbed correctly end to end — `graph.py:100` resolves `language or episode_config.language`, `graph.py:133` maps the code to a language name, `nodes.py` passes it to the template — it just lands in a template that ignores it. So `EpisodeProfile.language` looks supported and does nothing, which is worse than not having the field.

### Suggested fix

Add the bundled `{% if language %}` block to both app templates, immediately after `</speakers>`.

**The deeper fix is to stop shadowing.** The app templates have diverged in both directions: they *gained* the solo-speaker (`speakers|length == 1`) branches and the `<think>`-tag output rules, and they *lost* the language block. Every future library improvement to these prompts is invisible to the app. Either upstream the app-only features into `podcast-creator` and delete `open-notebook/prompts/podcast/`, or add a regression test that fails when a `{{ variable }}` present in the bundled template is missing from the app's.

Used a seeded speaker profile (`business_panel`) with a Gemini `voice_model` (`gemini-3.1-flash-tts-preview`) — which is what Vertex/Gemini provisioning and the docs recommend for such a deployment.

### How did it break?

After the transcript succeeded, every audio clip failed, was retried per clip, and the job died:

```
Google API error: Requested entity was not found.
```

`open_notebook/database/migrations/7.surrealql` seeds three speaker profiles whose `voice_id`s are **OpenAI** voice names: `nova` (lines 98, 118), `alloy` (104), `echo` / `shimmer` / `ash` (132, 138, 144). Migration 22 has since dropped the legacy `tts_provider` / `tts_model` strings those profiles carried, so a deployment whose live `voice_model` is Gemini inherits speaker profiles that can only work with OpenAI.

Proof that the model is fine and the *voice* is the problem — same model, same key, only the voice changed:

| Call | Result |
| --- | --- |
| `gemini-3.1-flash-tts-preview` + voice `echo` | **404 `Requested entity was not found.`** |
| `gemini-3.1-flash-tts-preview` + voice `Kore` | **200, 148 KB of audio** |
| `gemini-2.5-flash-preview-tts` + voice `echo` | 400 `Voice name echo is not supported. Allowed voice names are: achernar, achird, algenib, …` |

`SpeakerProfile.validate_speakers` (`open_notebook/podcasts/models.py:153-164`) checks only that a `voice_id` key *exists*, not that its value means anything for the profile's `voice_model`.

### Suggested fix

Three independent options:

1. **Validate the voice against the resolved TTS model before the run.** esperanto exposes the catalogue as `available_voices` on each TTS provider (30 voices for Gemini, `esperanto/providers/tts/google.py:117-302`). `validate_speakers` cannot see the model, but the podcast command can — rejecting an unknown voice there turns a five-minute failure into an instant, accurate one.
2. **Make the seed data provider-neutral or provider-correct** — or leave `voice_id` empty and let the UI fill it, rather than shipping values that are wrong for most deployments.
3. **Repair on upgrade**: a migration that rewrites `voice_id`s invalid for the profile's current `voice_model`, touching only invalid values so deliberate choices survive. Even better, offer the provider's voices as a dropdown in the speaker-profile form instead of a free-text field.

## 4. The error hint covers one case and misleads on the others

`commands/podcast_commands.py:367-373` appends a GPT-5 extended-thinking hint whenever the message contains `Invalid json output` or `Expecting value`. Useful, but it is the *only* hint, so the two most common real failures get either nothing (`Invalid speaker name`) or advice about the wrong provider (a truncated Gemini response gets told to switch to `gpt-4o`; `Requested entity was not found` on a Gemini deployment gets no hint at all).

Each of the defects above cost significantly more time to diagnose than to fix, purely because of this. Suggested additions:

| Message contains | Hint |
| --- | --- |
| `Invalid speaker name` | The model returned a placeholder or renamed a speaker. Names must match the speaker profile exactly. |
| `Requested entity was not found` / `Voice name … is not supported` | The speaker profile's `voice_id` is not valid for its TTS model. Check Settings → Speaker Profiles. |
| `Expecting value` after a long segment | The response may have been truncated — `max_tokens` for a transcript segment defaults to 5000 in `podcast-creator`; try fewer or shorter segments, or raise the profile's `max_tokens`. |

## Result downstream

Before, on a 6-segment Hebrew episode: English outline, then abort at segment 2/6 with `Invalid speaker name '...'`. After the fixes (including the `podcast-creator` retry fix): 6 segments, 42 Hebrew dialogue lines across all three speakers, MP3 produced, 474 s.

Happy to open a PR for the four items in this repo (templates, language block, pre-flight voice validation, error hints) with tests.

alefbt:
## Description

Rescoped per @lfnovo's review: this PR is now **only** the voice pre-flight. The templates and failure hints moved to #1334, which is the low-risk half and can land independently.

Audio is generated last, so a `voice_id` the TTS model doesn't accept fails only after the whole transcript has been generated and paid for — and the provider message rarely names the voice. Gemini's 3.x TTS preview answers an unknown voice with `404 Requested entity was not found.`, which reads like a missing model. Migration 7 seeds three profiles with OpenAI voice names (`nova`, `alloy`, `echo`, `shimmer`, `ash`), so any deployment whose voice model is Gemini inherits speakers that can only work with OpenAI. `validate_speakers()` only checks the `voice_id` key exists; nothing compares it to the model.

Same model, same key, only the voice changed:

| Call | Result |
| --- | --- |
| `gemini-3.1-flash-tts-preview` + voice `echo` | **404 `Requested entity was not found.`** |
| `gemini-3.1-flash-tts-preview` + voice `Kore` | **200, 148 KB of audio** |

`SpeakerProfile.validate_voices()` runs before generation and checks each speaker's voice — honoring per-speaker `voice_model` overrides — against esperanto's `available_voices`. It is deliberately conservative, because a false rejection is worse than the late failure it replaces: only a voice **another provider's catalogue claims** fails the run; a voice no catalogue knows is logged and allowed through, since those lists go stale (esperanto's OpenAI catalogue predates `ash`, which migration 7 itself seeds).

## Related Issue

Part of #1238. Templates and hints split out to #1334.

## Your review points

- **Rebased** onto current `main`. The `### Fixed` conflict you predicted did happen — the rebase produced duplicate `### Fixed`/`### Changed` blocks in `[Unreleased]`; merged by hand into the existing ones.
- **`STATIC_CATALOGUE_PROVIDERS` omitted azure and deepgram** — correct, both added. Deepgram resolves offline with 103 voices; Azure needed a placeholder `base_url` + `api_version` to construct, so the catalogue-only credentials are now per-provider (Google's constructor *rejects* an unexpected `endpoint`, so one shared dict couldn't serve both).
- **`vertex` burned the retry budget** — correct, removed. It can't construct without real Google credentials even when given a project id, so it never contributed an attribution and only spent attempts. A test now asserts every provider on that list actually resolves offline, so the next addition can't regress it.
- **`config_fingerprint` only sorted top-level keys** — correct, now `json.dumps(sort_keys=True, default=repr)`, which descends into nested mappings. Tests cover nested key-order stability, differing nested values, and unserializable values.

**On your dropdown question** — I think you're right that it's the better primary fix, and I'd rather do it as a follow-up than fold it in here. It prevents a bad id being *saved*, which this can't; but it also can't help the profiles migration 7 already seeded, which is every affected install today, and it spans `api/` + `frontend/`. My suggestion: land this as the backstop for existing rows, then a second PR adds `GET /api/models/{id}/voices` and turns the free-text field into a select. If you'd rather have the dropdown first and drop the pre-flight entirely, say so and I'll close this — the machinery here only exists because the id can currently be anything.

## Type of Change

- [x] Documentation update

## How Has This Been Tested?

- [x] Added new unit tests

**Test Details:** 747 tests pass, `ruff` and `mypy` clean. `tests/test_podcast_voice_validation.py` covers: Gemini's catalogue resolving offline and case-insensitively (`Kore` → `kore`); `echo` attributed to OpenAI and raising; `ash` warning but not raising; blank and whitespace-padded voices; an unavailable catalogue never blocking; per-speaker override resolution; per-account cache separation; one lookup per profile; a transient failure not disabling the rest of the pass; and the retry ceiling. No network or credentials needed.

**Still not smoke-tested end to end from this branch** — as you noted, and it's the reason this deserves its own review. The end-to-end Hebrew run in #1238 was on the reporting deployment. The realistic failure mode to look for is a false rejection: a valid voice that some provider's catalogue happens to list. `test_unattributable_voice_warns_instead_of_raising` pins the conservative path, but a real run on a non-Gemini deployment would be worth more than any of these tests.

## Design Alignment

- [x] Async-First for Performance

**Explanation:** The pipeline assumed OpenAI voices; a Gemini deployment paid for a full transcript before finding out. The pre-flight is `await`-ed and the only lookup that can block runs in a worker thread with a timeout.

### Code Quality
- [x] PEP 8, type hints, self-reviewed, commented where the reasoning isn't obvious
- [x] No new warnings or errors

### Testing
- [x] Tests that prove the fix; new and existing pass locally
- [x] `ruff check .` and `uv run python -m mypy .` clean

### Documentation
- [x] `docs/7-DEVELOPMENT/podcasts.md` gains a "Voice pre-flight" section: what fails the run, what only warns, and how the cache is keyed

### Database Changes
- [ ] No schema changes — this does **not** repair the seeded profiles; that was option 3 in the report and is still open

### Breaking Changes
- [ ] No breaking changes, though this can refuse a generation that previously started (and then failed late)


alefbt:
Second commit ([3579777](https://github.com/alefbt/open-notebook/commit/3579777)) addresses a review pass over the first:

- **Blank `voice_id` now fails up front.** `validate_speakers()` accepts a present-but-empty voice; it matches no catalogue and can't be attributed to any provider, so it hit the "unknown voice → warn and continue" branch and then failed during audio generation — the very late failure the pre-flight exists to prevent.
- **Catalogue lookups are memoized per pass** (`VoiceCatalogueCache`). The catalogue for a `(provider, model_name)` can't change between speakers of one profile, but each speaker re-fetched it and each unattributable voice re-enumerated all five static catalogues. On ElevenLabs/OpenRouter that was one network request per speaker, each able to run to the 10s timeout, before any work started.
- **Speaker names in the JSON example go through `tojson`.** A name containing a quote or backslash produced an invalid example — the one thing the example must never be.
- **The angle-bracket placeholders are gone**, replaced by fully written sample dialogue and segment values. `<the complete words this speaker says out loud>` was copyable in exactly the way this PR sets out to stop, contradicted the template's own "never emit placeholder content" rule, and would have been read aloud by the TTS engine. A copied example is now valid *and* speakable. Since the sample is hard-coded English, the language block labels it as such so it doesn't pull a non-English episode back toward English — happy to drop the sample entirely if you'd rather the prompt lean only on `format_instructions`.

Tests now parse the example with `json.loads` rather than string-matching it, and cover escaping, speakable dialogue, blank voices, the per-profile lookup count, and the warning on an unattributable voice. All five new assertions fail against the previous commit. 718 tests pass, ruff and mypy clean.

alefbt:
Fourth commit ([d51dbec](https://github.com/alefbt/open-notebook/commit/d51dbec)) — the cached-failure note is valid, fixed, plus one thing it led me to.

**Caching `None` did switch validation off for the rest of the profile.** I took a bounded-retry route rather than the suggested "only cache successes": caching nothing on failure means a genuinely dead endpoint pays `CATALOGUE_TIMEOUT_SECONDS` once per speaker, which is what the cache was added to prevent. So successes are memoized, and a failing key is retried up to `MAX_CATALOGUE_ATTEMPTS` (2) per pass — enough to shrug off a one-off timeout, capped at two waits instead of N. Happy to switch to the simpler unbounded retry if you'd rather.

**It also surfaced a real bug in my own `STATIC_CATALOGUE_PROVIDERS`.** That list is documented as "literal dict, no credentials, no network", but esperanto's Mistral provider paginates through `GET /audio/voices` — so every unplaceable voice sent a request (and collected a 401) to a provider the deployment may not even use. Mistral is out; `openai`, `google`, `vertex` and `xai` are genuinely literal dicts. The voice test file dropped from 2.7s to 0.5s once those calls stopped, which is how visible it was.

The attribution test now asserts the per-key bound rather than "no repeats", since a failing catalogue is legitimately retried once. 722 tests pass, ruff and mypy clean; both new tests fail against `4102c8b` (`DID NOT RAISE` for the transient case, and the attempt-count bound for the dead-endpoint case).

alefbt:
Rebased onto `9bd5f89` (no overlap — that commit only touches `.claude/**` and `.gitignore`) and addressed the seven findings in [e4a8d03](https://github.com/alefbt/open-notebook/commit/e4a8d03).

**Padded `voice_id` (models.py)** — correct, and my bug: I compared the `.strip()`-ed value while podcast-creator sends the *stored* one, so ` kore ` cleared the pre-flight and failed at TTS. Padding is now rejected with a message naming the value, rather than trimmed, so the check and the provider see the same string.

**Copied example undershoots `turns` / `num_segments`** — both examples are now labelled two-entry EXCERPTs and restate the count the response needs ("return at least {{ turns }} entries rather than the two shown"). I did *not* loop-render examples with `turns` (up to 10) or `num_segments` (up to 20) entries: that multiplies every prompt's token cost — the thing that made segment truncation a failure mode in the original report — to guard a hypothetical. Worth naming that these two findings, and the language one below, are consequences of the earlier review asking me to replace the angle-bracket placeholders with fully written sample values *so that a copy would be valid*. Each round trades one copy-hazard for another; this framing keeps the copy valid while making it read as partial.

**English sample in a non-English run** — unchanged beyond what's already there: an explicit bullet under the language block stating the sample is English only to show the structure and that every value must be in the target language. A localized sample isn't renderable without a translation step in the prompt layer. If you'd rather, the alternative is dropping the sample entirely when `language` is set and leaning on the prose plus `format_instructions` — one line, happy to switch.

**Docs wording** — fixed; the outline example has fixed segment values and the outline template never receives `speaker_names`, so the two examples are now described separately.

**Test assertions** — the attribution test now pins OpenAI's catalogue to exactly one lookup (the retry ceiling still covers Vertex, which legitimately retries), and the dialogue assertion checks non-empty text with no placeholder markers instead of requiring a trailing full stop, so `?` and `!` pass.

725 tests pass, ruff and mypy clean. The three new assertions fail against `d51dbec`.

lfnovo (changes requested):
Thanks for the thorough work here, and for sticking with the cubic loop. I verified the core finding: the local `prompts/podcast/*.jinja` shadow podcast-creator's bundled templates and dropped the `{{ language }}` block, so `EpisodeProfile.language` is inert today. That alone is worth landing.

Two asks before this can go in:

**1. Split into two PRs.** The bundle mixes three independent fixes with very different risk:
- Templates (`{{ language }}` restored, real JSON examples) + failure-specific hints in `podcast_commands.py`: low risk, mergeable as soon as they're rebased. Please open these as the first PR.
- Voice pre-flight (`open_notebook/podcasts/voices.py`, `SpeakerProfile.validate_voices()`): this is the only change that can refuse a generation that works today, and it hasn't been smoke-tested end to end from this branch. It deserves its own PR and its own review. Honest question for that PR: the cross-provider attribution, per-pass cache, retry budget and config fingerprint are a lot of machinery for "is this voice usable". A voice dropdown in the speaker profile form (populated from `available_voices`) would prevent the bad id from being saved in the first place. If you'd rather keep the pre-flight, a few smaller things: `STATIC_CATALOGUE_PROVIDERS` omits `azure` and `deepgram`, which are also literal dicts in esperanto; `vertex` in that list burns the retry budget on installs without `VERTEX_PROJECT`; and `config_fingerprint` only sorts top-level keys.

**2. Rebase on main.** The branch is behind and its `### Fixed` block will conflict with the one that landed in #1309.

One more: I'd take your offer to drop the hard-coded English sample dialogue from the prompts. Injecting English into non-English runs is a small new risk that the prose instruction only partially mitigates.

alefbt:
@lfnovo done, all four asks:

**1. Split.** Templates + hints are now [#1334](https://github.com/lfnovo/open-notebook/pull/1334) (low risk, standalone). This PR is rescoped to the voice pre-flight alone — title and description rewritten, branch rebuilt from current `main` so the diff is 6 files instead of 10.

**2. Rebased.** Your CHANGELOG prediction was right: the rebase produced duplicate `### Fixed`/`### Changed` blocks under `[Unreleased]` rather than a conflict marker. Merged by hand, one entry per PR.

**3. The three pre-flight points — all three confirmed:**
- `deepgram` resolves offline with 103 voices, `azure` with a placeholder `base_url` + `api_version`; both added. The catalogue-only credentials had to become per-provider: Google's constructor *rejects* an unexpected `endpoint`, so one shared dict couldn't serve both.
- `vertex` can't construct without real Google credentials even when handed a project id, so it never contributed an attribution and only burned attempts — removed. A test now asserts every provider on that list actually resolves offline.
- `config_fingerprint` now uses `json.dumps(sort_keys=True, default=repr)`, which descends into nested mappings; tests cover nested key-order stability, differing nested values and unserializable values.

**4. English sample dropped.** No sample is rendered at all when `language` is set (in #1334) — prose describes the structure and `format_instructions` carries the schema. Agreed it was a new risk for the exact runs the language fix is meant to help.

**On the dropdown:** I think you're right that it's the better primary fix, and I'd rather do it as a follow-up than fold it in. It stops a bad id being *saved*, which this can't — but it can't help the profiles migration 7 already seeded, which is every affected install today, and it spans `api/` + `frontend/`. Proposal: land this as the backstop for existing rows, then a PR adding `GET /api/models/{id}/voices` and turning the free-text field into a select. If you'd rather have the dropdown first and drop the pre-flight, say the word and I'll close this — the machinery only exists because the id can currently be anything.

747 tests pass here, 768 on #1334; ruff and mypy clean on both. Neither has been smoke-tested end to end from the branch — for this one the risk worth probing is a false rejection on a non-Gemini deployment.

alefbt:
@lfnovo — everything you asked for is in place. Summary with links, details in the [comment above](https://github.com/lfnovo/open-notebook/issues/1239#issuecomment-5554373781):

| Your ask | Where it landed |
| --- | --- |
| Split: templates + hints as their own PR | **[#1334](https://github.com/lfnovo/open-notebook/pull/1334)** — `{{ language }}` restored, copyable skeleton removed, `explain_generation_failure()` ([`6fb1510`](https://github.com/alefbt/open-notebook/commit/6fb1510)) |
| Split: pre-flight on its own | **This PR**, rebuilt from `main` and rescoped to 6 files ([`d3182ba`](https://github.com/alefbt/open-notebook/commit/d3182ba)); title and description rewritten |
| Rebase on main | Both branches are on current `main`. The `### Fixed` clash you predicted showed up as duplicate `Fixed`/`Changed` blocks under `[Unreleased]`, merged by hand — one entry per PR |
| `STATIC_CATALOGUE_PROVIDERS` missing azure/deepgram | Added — [`voices.py:57`](https://github.com/lfnovo/open-notebook/blob/d3182ba/open_notebook/podcasts/voices.py#L57). Needed per-provider placeholder credentials ([`catalogue_only_config()`, :74](https://github.com/lfnovo/open-notebook/blob/d3182ba/open_notebook/podcasts/voices.py#L74)): Google's constructor rejects an unexpected `endpoint`, Azure won't build without one |
| `vertex` burning the retry budget | Removed. It needs real Google credentials to construct even with a project id, so it only ever spent attempts; a parametrized test now asserts every provider on that list resolves offline |
| `config_fingerprint` top-level only | Now `json.dumps(sort_keys=True, default=repr)` — [`voices.py:140`](https://github.com/lfnovo/open-notebook/blob/d3182ba/open_notebook/podcasts/voices.py#L140). Tests cover nested key-order stability, differing nested values, unserializable values |
| Drop the hard-coded English sample | Done in #1334: no sample renders when `language` is set; prose describes the structure, `format_instructions` carries the schema |

747 tests pass here, 768 on #1334; ruff and mypy clean on both. #1334 is independent — it doesn't touch `voices.py` or `models.py` and can go in without this one.

Two things still open on your side: the dropdown-vs-pre-flight call (my proposal is in the comment above — land this as the backstop for the already-seeded profiles, dropdown as a follow-up, and I'll close this instead if you prefer the dropdown alone), and neither PR has been smoke-tested end to end from the branch.

fady-mz:
I ran current head `e1e96dc` against the frozen zero-provider-spend FinOps workload for issue #1238. The scoped behavior is sound: 42 focused tests passed, Ruff and MyPy passed, and a command-boundary harness confirmed an incompatible voice stops before profile enumeration, episode persistence, output-directory creation, configuration, or `create_podcast()`.

Two points look worth addressing before merge:

1. The preflight currently calls `available_voices` for every active provider in `get_known_voice_ids()`. For HTTP-backed implementations such as ElevenLabs and OpenRouter, that adds up to two network attempts / 20 seconds before generation. The frozen quality guard was stricter: only inspect explicitly proven local catalogues (the two Google Gemini TTS model IDs) and fail open without probing unknown, dynamic, incomplete, or network-backed catalogues. Could the policy return unknown before constructing those providers, or otherwise prove the preflight itself performs no HTTP?
2. The PR tests exercise `validate_voices()` but not its position in `generate_podcast_command`. A regression test should assert an incompatible scoped voice produces zero `repo_query`, `PodcastEpisode.save`, output-directory, `configure`, and `create_podcast` calls. The current head passes that check in my external harness; keeping it in-repo would protect the cost-saving boundary.

Minor documentation drift: the module docstring still lists Vertex and Mistral among hard-coded/free catalogues, while `STATIC_CATALOGUE_PROVIDERS` intentionally excludes both.

The full Windows run was 743 passed / 4 failed; all four failures are in unchanged POSIX path/proxy tests, not this diff.

Issue #947, opened 2026-06-21
[Design] Model & Provider Configuration (advanced params, presets, health)


lfnovo:
**[Umbrella]** Consolidates the scattered per-model / per-provider configuration requests so we design the architecture **once**, then ship it in phases. Design ratified 2026-07-21 (discussion summary below).

Consolidates (all closed, folded here): #442, #682, #831, #696, #879, #883, #689. Related resilience eixo: #1156, #1157, #548.

---

### 1. Config hierarchy — phased, provider-level first
Layered merge, most-specific wins, resolved at `ModelManager.get_model`.

- **Phase 1 (now):** `L1 connection` (`Credential.config` FLEXIBLE bag) → `L3 runtime` (per-use overrides). Provider-level config first — get comfortable with it before per-model.
- **Phase 2 (later):** `L2 per-model` override layer (leading candidate: a name-keyed `model_overrides` bag inside the connection config — zero `model`-table migration, survives bulk re-discovery). **Param presets (#696) land here too** — a preset is "just another override form", so it depends on the provider layer being settled, and must be **valid cross-provider** (reasoning models reject `temperature`, ranges differ, etc.).

### 2. Parameter surface — two tiers, provider-declared
- **Universal tier** (`temperature`/`top_p`/`max_tokens`) — every language provider; per-provider rule annotations (Anthropic `top_p`↔`temperature`).
- **Provider-specific tier** — each `ProviderSpec` declares `tunables`: `num_ctx`/`keep_alive` (Ollama), `extra_body` (OpenAI-compatible). UI renders the union. Generalizes today's hardcoded `isOllama` conditional.
- Persistence: existing `CONFIG_EXTRAS` FLEXIBLE-bag pattern — no per-param migration.

### 3. Model types — esperanto modalities only
No pseudo-types (coding/OCR). Type = what `AIFactory.create_*` can build. Reranker is the only genuinely new modality (#1087, own card); OCR belongs to the extraction-engine axis (#1132). #689 closed on this basis.

### 4. Health — probe now, taxonomy upstream
Active probe reuses `connection_tester`. The passive error taxonomy **consumes esperanto#227** (normalized exception hierarchy) — ON does not hand-roll a classifier. Behaviors (fallback/prune/warning) live in #1156/#1157/#1155.

### 5. Recommendations (#879) — documentation, not a feature
[...]

Issue #830, opened 2026-06-02
feat(chat): expose temperature/top_p controls


lfnovo:
## Context (Phase 3 — future)

Esperanto 2.21.0 added per-call `max_tokens` / `temperature` / `top_p` overrides on `chat_complete()` / `achat_complete()` for **all** LLM providers. Open Notebook currently only threads `max_tokens` (`open_notebook/ai/provision.py`); `temperature` and `top_p` are not exposed.

## Possible scope

- Allow `temperature` / `top_p` overrides through `provision_langchain_model()` / `ModelManager`.
- Surface controls in chat / model settings UI + i18n.
- Note: for Anthropic, `top_p` is dropped when `temperature` is set (provider rule).

## Notes

Future / not scheduled. Part of the esperanto 2.22.0 follow-ups.

lfnovo:
Promoted to `ready` as Phase 1 of the #947 umbrella (L3 runtime controls).

**Scope:** expose `temperature` / `top_p` / `max_tokens` as per-use runtime controls in chat (the threading mechanism already exists in `provision.py` via kwargs → esperanto `config`; there is simply no caller/UI today). These are the **universal tier** — they apply to every language provider. Carry the provider rule: **Anthropic drops `top_p` when `temperature` is set** → validate/warn. Named param presets over these values are deferred to Phase 2 (#947 §3a).

lfnovo:
Removing the `ready` label. The design for this card is settled in the umbrella (#947) and nothing here is being reconsidered — but `ready` is a promise that someone is about to build it, and this card has had no builder since it was written. Keeping it there recreates the stale backlog the contribution model ([#1212](https://github.com/lfnovo/open-notebook/discussions/1212)) is meant to avoid. It stays tracked in #947 and goes back to `ready` the moment someone picks it up — say so here if that's you.

bferanmi806-sketch:
I’d like to take this one. I’m reading #830 as the Phase 1 L3 runtime controls from #947: temperature, top_p, and max_tokens for chat, with unset values deferring to lower config layers rather than becoming persistent per-model settings.

One thing I want to confirm before I start: for max_tokens, should an unset runtime value preserve the current chat 8192 fallback only after any L1 connection default has had a chance to apply, rather than always sending 8192 as an L3 override?

BZoboki:
[...]

PR #1247, opened 2026-08-06
fix: make Ask output token budget configurable

### LINKED ISSUE #1221: [Bug] Ask (Q&A) answers are silently truncated due to hardcoded max_tokens=2000 in graphs/ask.py — especially visible in non-English languages
## Summary

`POST /api/search/ask/simple` returns HTTP 200 with an answer that is cut off mid-sentence, with no error indication to the caller. Root cause appears to be that `open_notebook/graphs/ask.py` hardcodes `max_tokens=2000` in all three LLM calls of the ask graph, with no config/env override path. This is the same pattern already fixed for transformations (#565 -> #568) and podcasts (#639 -> #982/#992/#991/#1197), but the ask/Q&A graph appears to have been missed.

Because non-English languages (e.g. Japanese) often consume substantially more tokens for the same amount of content, this cap is hit far more easily in those languages, making the bug much more visible for non-English-speaking users while being harder to notice for English-only maintainers/testers.

## Environment

- Docker image: `lfnovo/open_notebook:v1-latest` (rolling tag; exact commit could not be pinned down — Docker Hub shows the same digest as `1.14.0` for `linux/amd64` as of 2026-07-25, but this was not independently verified beyond the tag page)
- Chat model: Anthropic `claude-sonnet-5`
- Embedding model: Ollama `nomic-embed-text`
- Confirmed independently that the same hardcoding exists in the current `main` branch on GitHub (see "Root cause" below), so this is not specific to our image/version.

## Steps to Reproduce

1. Set up a notebook with sources in Japanese (or any token-dense non-English language).
2. Call `POST /api/search/ask/simple` with a Japanese question.
3. Repeat with a few different Japanese questions against the same notebook.

## Observed Behavior

- HTTP status is 200 (no error surfaced).
- In our testing, 2 out of 3 Japanese questions returned answers cut off mid-sentence / mid-token. One answer ended abruptly mid-sentence; another ended in the middle of what looks like a source ID reference.
- This is consistent with hitting a hard output-token cap during generation, though we did not inspect `finish_reason`/`stop_reason` from the underlying provider response to confirm this directly — flagging this as something maintainers may want to check when reproducing.
- There is no warning, error, or truncation indicator returned to the API caller. The user has no way to know the answer is incomplete.

## Root Cause (verified against upstream `main`)

In `open_notebook/graphs/ask.py` (main branch, as of 2026-07-25), all three `provision_langchain_model(...)` calls pass a hardcoded `max_tokens=2000`, with no config or environment variable to override it:

- `call_model_with_messages()` — the strategy-generation call (structured/JSON output)
- `provide_answer()` — per-search-result answer generation
- `write_final_answer()` — final answer synthesis (the one most likely responsible for the user-visible truncation)

We were not able to obtain a fully reliable exact line number for these three call sites — two different retrieval methods against the same file gave inconsistent line numbers (apparently due to a rendering artifact, not a real code difference), so we're intentionally not quoting a specific line number here to avoid pointing at the wrong line. They are easy to find with:

    grep -n "max_tokens=2000" open_notebook/graphs/ask.py

We only confirmed the hardcoding exists — we have not confirmed that all three call sites are equally responsible for the truncation we observed. The first call (`call_model_with_messages`) requests structured/JSON output, so hitting the cap there would more likely surface as a JSON-parsing error rather than a silently truncated prose answer. The truncation we actually observed looks more consistent with the `provide_answer` / `write_final_answer` stages.

## Why This Is More Visible in Non-English Languages

For a given amount of semantic content, tokenizers used by most LLM providers typically require more tokens to represent Japanese (and several other non-English/non-Latin-script languages) than to represent equivalent English content. A fixed 2000-token output cap that is rarely hit in English can be hit routinely in Japanese, especially for longer or more detailed answers. This means the same codebase can look "fine" in English-language testing while silently degrading for non-English users.

## Precedent

This exact class of bug (hardcoded `max_tokens` causing silent truncation) has already been reported and fixed twice in this repo for other code paths:

- #565 — Transformation output truncated due to hardcoded `max_tokens` (fixed in #568)
- #639 — Podcast generation hardcoded `max_tokens` (fixed via #982/#992, further exposed in UI via #991/#1197)

The ask/Q&A graph (`graphs/ask.py`) appears to be the one remaining major code path with the same hardcoding.

Separately, #947 (Model & Provider Configuration design, ratified 2026-07-21) lists a universal parameter tier including `max_tokens` as part of its Phase 1 scope, and #830 notes that `provision_langchain_model()` already supports per-call `max_tokens` overrides via Esperanto >=2.21.0. This suggests the underlying plumbing to fix this already exists; it's a matter of wiring `graphs/ask.py` to use a configurable value rather than a literal `2000`.

## Suggested Fix (any of the following, not mutually exclusive)

1. Make `max_tokens` for the ask graph configurable via settings/env var (consistent with how #991/#1197 exposed it for podcasts).
2. Raise the hardcoded default meaningfully (e.g. similar to how #568 raised the transformation default), acknowledging that a single number still won't be right for all languages/models.
3. At minimum, detect when the model output was cut off due to the token cap (e.g. via `finish_reason`/`stop_reason` if available from the provider) and surface a warning to the API caller / UI instead of silently returning HTTP 200 with an incomplete answer.

## Open Questions for Maintainers

- Can you confirm via `finish_reason`/`stop_reason` whether the observed truncation is in fact due to hitting the 2000-token cap, as opposed to some other cause?
- Does this reproduce with providers other than Anthropic? We have only tested with `claude-sonnet-5`.

## Notes on What We Could Not Verify

- We could not pin down the exact commit corresponding to our `lfnovo/open_notebook:v1-latest` image (rolling tag).
- We did not verify this against a from-source / non-Docker install.
- We did not search GitHub Discussions (only Issues/PRs) for prior reports.

bferanmi806-sketch:
## Description

Make the Ask/Q&A output token budget configurable across the strategy, intermediate-answer, and final-answer model-generation stages so reasoning models have sufficient budget before thinking content is removed.

- Structured strategy generation, intermediate answers, and final answers all use `OPEN_NOTEBOOK_ASK_MAX_TOKENS`.

- The default is `8192` tokens.
- Malformed, zero, and negative values safely fall back to `8192` with a warning.
- Added focused regression tests and configuration documentation.
- No API, frontend, provider, dependency, or database behavior changed.

## Related Issue

Fixes #1221

## Type of Change

- [ ] Breaking change (bug or feature that would cause existing functionality to not work as expected)
- [x] Documentation update
- [ ] Code refactoring (no functional changes)
- [ ] Performance improvement

## How Has This Been Tested?

- [ ] Tested locally with Docker
- [x] Added new unit tests
- [ ] Existing tests pass (`uv run pytest`)
- [ ] Manual testing performed (describe below)

**Test Details:**

- `uv sync` - completed successfully.
- `uv run pytest tests/test_ask_token_budget.py -v` - 9 passed, 1 dependency deprecation warning.
- `uv run pytest tests/ -q` - 612 passed, 3 failed, 49 errors, 2 dependency warnings. The same three Windows path/proxy-environment failures and 49 temp-directory/Windows-permission errors documented for this PR remain; no Ask-related tests failed.
- `uv run ruff check .` - passed.
- `uv run ruff format --check open_notebook/graphs/ask.py tests/test_ask_token_budget.py` - 2 files already formatted.
- `uv run python -m mypy .` - no issues found in 131 source files.
- `git diff --check` - passed.

## Design Alignment

- [ ] Privacy First
- [ ] API-First Architecture

**Explanation:**

The change adds one narrow, process-level configuration helper instead of expanding the request/API contract. It preserves provider-agnostic model selection and passes the budget through the existing per-call Esperanto configuration path.

### Code Quality
- [ ] My code follows TypeScript best practices (Frontend; no frontend changes)
- [ ] I have added JSDoc comments where appropriate (no TypeScript changes)
- [x] I have commented my code, particularly in hard-to-understand areas (the helper is self-documenting and has no complex undocumented logic)

### Testing
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes (focused tests pass; unrelated baseline failures are documented above)
- [x] I ran linting: `make ruff` or `ruff check . --fix`
- [x] I ran type checking: `make lint` or `uv run python -m mypy .`

### Documentation
- [x] I have updated the relevant documentation in `/docs` (if applicable)
- [ ] I have updated the API documentation (no API changes)
- [x] I have added comments to complex logic (no complex logic was added)

### Database Changes
- [ ] I have created migration scripts for any database schema changes (no database changes)
- [ ] Migration includes both up and down scripts
- [ ] Migration has been tested locally

### Breaking Changes
- [ ] I have documented the migration path for users
- [ ] I have updated MIGRATION.md (if applicable)

## Screenshots (if applicable)

Not applicable; this is a backend-only change.

## Additional Context

All three Ask model-generation stages use the same cached `OPEN_NOTEBOOK_ASK_MAX_TOKENS` value: structured strategy generation, fan-out intermediate answers, and final-answer synthesis. The existing structured JSON response format and model-selection behavior remain unchanged.

Before submitting, please verify:

- [x] This PR addresses approved Issue #1221 claimed by `bferanmi806-sketch`
- [x] I have not included unrelated changes
- [x] My PR title follows conventional commits format: `fix: make Ask output token budget configurable`

---

**Thank you for contributing to Open Notebook!**

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1247?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -70,6 +70,12 @@ SURREAL_DATABASE=open_notebook
 # to match.
 # OPEN_NOTEBOOK_MAX_UPLOAD_SIZE_MB=100
 
+# Maximum output tokens for Ask/Q&A intermediate and final answers (default: 8192).
+# The structured strategy-generation step remains fixed at 2000. Invalid or
+# non-positive values fall back to 8192. Restart the API process/container after
+# changing this value. Higher values can increase latency and token usage.
+# OPEN_NOTEBOOK_ASK_MAX_TOKENS=8192
+
 # Security
 # Password to protect this Open Notebook instance. Supports Docker secrets
 # via OPEN_NOTEBOOK_PASSWORD_FILE. Auth is fully disabled if unset.
--- a/.env.example
+++ b/.env.example
@@ -70,8 +70,8 @@ SURREAL_DATABASE=open_notebook
 # to match.
 # OPEN_NOTEBOOK_MAX_UPLOAD_SIZE_MB=100
 
-# Maximum output tokens for Ask/Q&A intermediate and final answers (default: 8192).
-# The structured strategy-generation step remains fixed at 2000. Invalid or
+# Maximum output tokens for all Ask/Q&A model-generation stages: structured
+# strategy, intermediate answers, and final answers (default: 8192). Invalid or
 # non-positive values fall back to 8192. Restart the API process/container after
 # changing this value. Higher values can increase latency and token usage.
 # OPEN_NOTEBOOK_ASK_MAX_TOKENS=8192
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Community contribution intake now separates exploration from execution: feature requests, product/design/architecture ideas and contribution proposals start in GitHub Discussions, while Issues are reserved for reproducible bugs and maintainer-approved work items. The Issue chooser routes contributors accordingly, a structured Ideas Discussion form starts from user goals and outcomes, and the contributor/maintainer docs plus PR template now describe the Discussion → Issue → PR graduation path (#1204).
 - Release image gate gained a `probe` scenario (`make release-test` runs it as part of `all`): container-level checks that a Python test suite can't cover because they depend on the shipped image's process supervision — `OPEN_NOTEBOOK_WORKER_MAX_TASKS` reaching the in-image worker (the supervisord `sh -c` expansion), and the worker surviving startup with `HTTP_PROXY` set while a user's `NO_PROXY` value is preserved (the internal SurrealDB websocket not being tunneled). Both were manual probes during the v1.14.0 release; they now run automatically. Release-process docs gained the post-tag re-cut sequence and a note on never leaving the version bump uncommitted (v1.14.0 retro)
 
+### Fixed
+- Ask/Q&A intermediate and final answers now use a configurable 8192-token output budget via `OPEN_NOTEBOOK_ASK_MAX_TOKENS`, while structured strategy generation remains fixed at 2000; invalid or non-positive values safely fall back to the default (#1221)
+
 ## [1.14.0] - 2026-07-20
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Release image gate gained a `probe` scenario (`make release-test` runs it as part of `all`): container-level checks that a Python test suite can't cover because they depend on the shipped image's process supervision — `OPEN_NOTEBOOK_WORKER_MAX_TASKS` reaching the in-image worker (the supervisord `sh -c` expansion), and the worker surviving startup with `HTTP_PROXY` set while a user's `NO_PROXY` value is preserved (the internal SurrealDB websocket not being tunneled). Both were manual probes during the v1.14.0 release; they now run automatically. Release-process docs gained the post-tag re-cut sequence and a note on never leaving the version bump uncommitted (v1.14.0 retro)
 
 ### Fixed
-- Ask/Q&A intermediate and final answers now use a configurable 8192-token output budget via `OPEN_NOTEBOOK_ASK_MAX_TOKENS`, while structured strategy generation remains fixed at 2000; invalid or non-positive values safely fall back to the default (#1221)
+- Ask/Q&A structured strategy generation, intermediate answers, and final answers now use the configurable 8192-token output budget from `OPEN_NOTEBOOK_ASK_MAX_TOKENS`; invalid or non-positive values safely fall back to the default (#1221)
 
 ## [1.14.0] - 2026-07-20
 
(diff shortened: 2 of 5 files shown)


lfnovo:
Thanks for this. The change itself is sound, and I'm parking it deliberately rather than for quality reasons: we're consolidating how max_tokens (and the other model tunables) get configured under #947, with #830 moving them into Settings instead of one environment variable per workflow. I'd rather not add another env var now and deprecate it a release later. I'll revisit this PR once #830 lands; if it ends up being the natural place for this knob, we'll fold it in with credit.

lfnovo:
Closing now that #1323 landed on `main` for #1221: the three Ask stages share the 8192 budget your PR defaulted to, plus the empty-strategy and truncated-thinking handling. I'm not taking the environment variable itself, for the reason in the parking note above: model tunables (max tokens included) are being consolidated under #947, with #830 as the card that puts them in Settings per model instead of one variable per workflow. Your validation and test approach was solid, and the tests in #1323 borrow from the same idea. Thank you for the work, and if #830 interests you, it's open for a builder.

PR #1286, opened 2026-08-29
fix: open the note editor from Write Note

### LINKED ISSUE #1252: [Bug]: "Write Note" button in a notebook does nothing (no modal, no console error, no network request)

Open an existing notebook (with at least one source already added); click the "Write Note" / "New Note" button in the notebook's notes section; observe that nothing happens.

### How did it break?

Expected: a modal/editor should open to create a new note. Actual: clicking the button has zero effect — no modal appears, no error message is shown, no error is logged to the browser console, and no network request is fired. Reproduced twice in a row. Other note-related and source-related flows in the same notebook work normally.

### Logs or Screenshots

No console errors and no entries in the Network tab at all when clicking the button — confirmed via browser DevTools. This suggests the click handler for the "Write Note" button is either not wired up or is silently failing before making any request.

### Open Notebook Version

v1-latest (Docker)

### Environment

- OS: macOS (Docker via OrbStack) - Browser: Google Chrome - Deployment: docker compose, image lfnovo/open_notebook:v1-latest

### Additional Context

Found while doing a full manual functionality pass on a fresh install. Everything else in the Notebooks flow (create/delete notebook, add sources, chat/search gating without an embedding model, transformations playground) works as expected — this button is an isolated dead click. Happy to provide a screen recording if useful.

### Contribution

mvanhorn:
## Description

Replace the coupled create/edit open-state calculation in `NotesColumn` with one explicit editor-open state and one optional selected note, following the controlled editor pattern already used by `TransformationsList`. Route both the Write Note button and note-row editing through a small production handler that clears or sets the selected note before opening the existing `NoteEditorDialog`, and mirror `onOpenChange` directly while clearing selection on close. Add a `NotesColumn` interaction test covering the create and edit entry points.

## Scope: refactor and test coverage, not a fix for #1252

This PR originally claimed to fix #1252. It does not, and the description has been corrected.

As @lfnovo showed in review, `NotesColumn.test.tsx` passes unmodified against pre-PR `main`: React batches the `setEditingNote(null); setShowAddDialog(true)` pair, so `open` already resolves to `true` on click under jsdom. The test therefore does not reproduce #1252, and the claim that the entry point derives open state incorrectly is not borne out by the code on `main`.

What remains is still worth landing on its own terms: one explicit state instead of two coupled booleans, consistency with the existing `TransformationsList` pattern, and interaction coverage for a column that previously had none.

#1252 stays open. Its reporter saw no modal, no console error, and no network request in Chrome against the `v1-latest` Docker image, which points at something jsdom cannot observe (an overlay or z-index swallowing the click, or an environment-specific condition that may since have changed) rather than at this state calculation.

Refs #1252

## Related Issue

Refs #1252

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [x] Code refactoring (no functional changes)
- [ ] Performance improvement

## How Has This Been Tested?

`NotesColumn.test.tsx` covers, at the mocked editor boundary: the create path receives the current notebook ID with no selected note; the dialog opens on click, closes via `onOpenChange(false)`, and reopens without stale edit state; and the edit path still opens with the clicked note, with selection cleared afterwards.

These assert the refactor preserves behavior. They are not a regression test for #1252, and they pass on pre-PR `main` as well.

Not verified: no browser reproduction against `main` was attempted, so this PR makes no claim about #1252's actual cause.

## Design Alignment

- [ ] Privacy First
- [ ] API-First Architecture
- [ ] Multi-Provider Flexibility

**Explanation:** One explicit editor state replaces two coupled booleans whose interaction had to be reasoned about at every call site.

### Code Quality

- [x] My code follows TypeScript best practices (Frontend)

### Testing

- [x] I have added interaction tests covering the refactored entry points
- [ ] I have added tests that prove a fix is effective (not applicable: this PR is not a fix)
- [ ] I ran linting: `make ruff` or `ruff check . --fix`
- [ ] I ran type checking: `make lint` or `uv run python -m mypy .`

### Documentation

- [ ] I have updated the relevant documentation in `/docs` (not applicable: no user-facing behavior change)

### Database Changes

- [ ] Not applicable: this PR contains no schema changes and no migrations

### Breaking Changes

## Additional Context

Nothing beyond what is described above.

- [x] This PR addresses an approved Issue assigned to me, **or** it's a small obvious fix
- [x] My PR title follows conventional commits format

---

AI was used for assistance.


lfnovo:
Thanks for this. The refactor to a single `editorOpen` + optional `editingNote` is cleaner than the two coupled booleans, and the new `NotesColumn.test.tsx` is a welcome addition.

One thing gives me pause: I dropped your test file, unmodified, onto current main (the pre-PR `NotesColumn.tsx`) and all three tests pass there. React batches the `setEditingNote(null); setShowAddDialog(true)` pair, so `open` already resolves to `true` on click in jsdom. That means the test doesn't reproduce #1252, and the PR's claim that the entry point derives open state incorrectly isn't borne out by the code on main.

My guess is the real cause of #1252 is something jsdom can't see, such as an overlay or z-index swallowing the click in a real browser, or something environment-specific that has since changed. Could you try to reproduce #1252 on current main in a browser and, if it still reproduces, capture what actually blocks the click? If it doesn't reproduce anymore, I'm still open to landing this as a clarity refactor plus tests, just with the description reworded so the changelog doesn't claim a fix we can't demonstrate.

mvanhorn:
You're right, and thanks for actually running the test against main rather than taking my word for it. The batching explanation holds up: `open` resolves true on click under jsdom either way, so the test proves the refactor is behavior-preserving and nothing more.

I've reworded the description to drop the causal claim and the `Fixes` link. It's now scoped as a clarity refactor plus interaction coverage, with a note that #1252 stays open and likely has a cause jsdom can't see, along the lines you suggested.

I haven't tried the browser reproduction, so I'm not claiming anything about what actually blocks the click in #1252. Happy to leave that to whoever can get on the `v1-latest` Docker image the reporter was running, or to pick it up separately if it's still live.

Issue #1290, opened 2026-08-30
API password auth: a non-latin-1 password can never authenticate


Ro-28:
## Problem

A password containing characters outside latin-1 can never authenticate over HTTP. The result is a permanent 401 with no way to succeed and no error explaining why. Pure-ASCII passwords are unaffected, which is why this has gone unnoticed.

ASGI carries raw header bytes and Starlette decodes header values as latin-1, so the credential string reaching the middleware is the latin-1 decoding of whatever bytes the client sent. `api/auth.py:71-73` then re-encodes that string as UTF-8 before comparing it against the UTF-8 bytes of the configured password:

```python
if not secrets.compare_digest(
    credentials.encode("utf-8"), self.password.encode("utf-8")
):
```

For `pässwörd-中文` a UTF-8 client sends bytes that decode to mojibake, and no value the client can send will match.

## Fix

`api/auth.py:71` — recover the wire bytes instead of re-encoding the mojibake:

```python
if not secrets.compare_digest(
    credentials.encode("latin-1"), self.password.encode("utf-8")
):
```

Stay in bytes rather than round-tripping back to `str`: a `.decode("utf-8")` on attacker-supplied garbage would raise and turn a 401 into a 500. A string decoded from latin-1 is always latin-1-encodable, so no `try/except` is needed.

Behaviour change worth a CHANGELOG line: today a client sending the *latin-1* bytes of a latin-1-representable password (`pässwörd`) authenticates, because Starlette decodes it back to the original string. That path stops working, in exchange for the UTF-8 path that modern clients actually use. Hard to hit in practice — httpx (and therefore TestClient) rejects non-ASCII header values, and curl/browsers emit UTF-8.

## Tests

There is no coverage for `api/auth.py` today. Add `tests/test_auth.py` with `TestPasswordAuthMiddleware`, `TestNoPasswordConfigured` and `TestNonAsciiPassword`. The non-ASCII case needs a raw-ASGI helper that drives the app directly, because TestClient cannot send non-ASCII header bytes.

## Acceptance criteria

- A deployment with `OPEN_NOTEBOOK_PASSWORD=pässwörd-中文` authenticates when the client sends the password UTF-8-encoded.
- ASCII passwords keep working unchanged.
- Invalid UTF-8 in the Authorization header returns 401, not 500.
- `tests/test_auth.py` covers the three classes above.
- CHANGELOG entry under Unreleased → Fixed, noting the latin-1 client behaviour change.

[...]

PR #1026, opened 2026-07-08
chore: remove dead auth helper and fix stale default-password docs

kwp3:
\`check_api_password()\` (an unused HTTPBearer-based dependency, superseded by \`PasswordAuthMiddleware\`) and its now-unused imports are dead code - nothing calls it. Removed.

Docs across \`api/CLAUDE.md\`, \`docs/3-USER-GUIDE/api-configuration.md\`, \`docs/5-CONFIGURATION/security.md\`, \`docs/7-DEVELOPMENT/security.md\`, and \`docs/SECURITY_REVIEW.md\` still described a hardcoded default password (\`open-notebook-change-me\`) that \`PasswordAuthMiddleware\` doesn't actually have - if \`OPEN_NOTEBOOK_PASSWORD\` is unset, auth is fully disabled instead. Updated to match actual behavior.

Docs-only + dead-code cleanup, not a new fix. **Stacked on #1002-#1009** (unmerged). Full test suite passes (222/222) on this branch.

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1026?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #1291, opened 2026-08-30
fix(auth): constant-time password compare and UTF-8 credential decoding

Ro-28:
Fixes the two authentication issues described in #1290, plus a docstring correction, and adds test coverage for a module that previously had none.

## Changes to `api/auth.py`

**Constant-time comparison.** Both password checks now use `hmac.compare_digest` on encoded bytes instead of `!=` on strings, so response timing no longer varies with how much of the password prefix was correct.

**UTF-8 credential decoding.** A new `_credential_bytes()` helper re-encodes the incoming credential with latin-1 to recover the exact bytes the client put on the wire, and compares those against the configured password's UTF-8 bytes. A credential that is not latin-1 encodable did not arrive over HTTP, so it falls back to UTF-8 for direct callers of `check_api_password`.

**Docstring.** `PasswordAuthMiddleware` now says it is disabled entirely when `OPEN_NOTEBOOK_PASSWORD` is unset, which is what the code actually does.

## Behavior change worth reviewing

Clients must send UTF-8 for non-ASCII passwords. Previously, a client sending latin-1 wire bytes for a latin-1-representable password such as `pässwörd` would succeed; it now returns 401.

**Pure-ASCII passwords are unaffected**, since their latin-1 and UTF-8 encodings are identical — so this should be invisible to essentially every existing deployment. But it is a real change and I would rather flag it than bury it. If you would prefer to accept both encodings for backward compatibility, that is an easy adjustment.

## Tests

Adds `tests/test_auth.py` with 61 tests covering the middleware, the `check_api_password` dependency, and `GET /auth/status`:

- correct, wrong, and prefix passwords
- missing, empty, and malformed `Authorization` headers
- the default excluded paths, exact-match path exclusion, and custom `excluded_paths`
- `OPTIONS` preflight, and non-GET methods
- init-time (middleware) vs call-time (dependency) password reads
- wire-encoding behavior for non-ASCII passwords
- `/auth/status` never leaking the password value

The tests build a bare FastAPI app rather than importing the production app, so they need no database and no network. Full suite: 283 passed.

---

https://claude.ai/code/session_01KPzTa2pWc3FenutyvuqUXu

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1291?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/api/auth.py
+++ b/api/auth.py
@@ -1,3 +1,4 @@
+import hmac
 from typing import Optional
 
 from fastapi import Depends, HTTPException, Request
@@ -10,10 +11,26 @@ from starlette.types import ASGIApp
 from open_notebook.utils.encryption import get_secret_from_env
 
 
+def _credential_bytes(credential: str) -> bytes:
+    """
+    Recover the raw bytes a client sent for an Authorization credential.
+
+    ASGI servers decode header values as latin-1, so re-encoding with latin-1
+    restores the exact bytes on the wire and lets a UTF-8 password (including
+    characters outside latin-1) match. A credential that is not latin-1
+    encodable did not arrive over HTTP -- it came from a direct call -- so
+    fall back to UTF-8 for those callers.
+    """
+    try:
+        return credential.encode("latin-1")
+    except UnicodeEncodeError:
+        return credential.encode("utf-8")
+
+
 class PasswordAuthMiddleware(BaseHTTPMiddleware):
     """
     Middleware to check password authentication for all API requests.
-    Always active with default password if OPEN_NOTEBOOK_PASSWORD is not set.
+    Disabled entirely if OPEN_NOTEBOOK_PASSWORD is not set.
     Supports Docker secrets via OPEN_NOTEBOOK_PASSWORD_FILE.
     """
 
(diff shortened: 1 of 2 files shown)


lfnovo (changes requested):
Thanks for this, and for the careful write-up in the description. The branch is stale against main, and that changes the scope quite a bit:

- `secrets.compare_digest` already landed in #1003 (`api/auth.py`), so the `hmac.compare_digest` half is a no-op rename that now conflicts.
- The docstring was corrected in #1026.
- `check_api_password` was removed in #1026 as dead code. The PR re-adds a fix to it, and `tests/test_auth.py` imports it, so the whole test module fails to collect on main.

What's still new and correct is the latin-1 recovery of the wire bytes so non-ASCII UTF-8 passwords can authenticate. I confirmed the bug on main: `credentials.encode("utf-8")` against a latin-1-decoded header can never match a UTF-8 password. Once rebased, the `except UnicodeEncodeError` fallback becomes unreachable (a latin-1-decoded string is always latin-1-encodable), so `_credential_bytes` can collapse to a single `encode("latin-1")` or be inlined.

Could you rebase and re-scope to: the latin-1 fix in `dispatch`, the `TestNonAsciiPassword` class, and the `TestPasswordAuthMiddleware` / `TestNoPasswordConfigured` coverage (genuinely new, we have no auth tests today)? Drop the `check_api_password` classes. Please add a one-line CHANGELOG entry under Unreleased → Fixed. Keep in mind AGENTS.md frames this middleware as a dev default rather than production hardening, so smaller is better here.

PR #1293, opened 2026-08-30
fix(database): batch large inserts using one connection

gblue1223:
## Description

Large documents can produce thousands of embedding records with 3072-dimensional vectors. Sending them in one INSERT can exceed WebSocket frame limits and reset the connection. `repo_insert` now inserts batches of 50 records while opening one database connection for the entire call.

Preserves result order, record-ID conversion, and duplicate-batch handling. Empty input opens no connection; non-positive batch sizes are rejected.

The Docker Compose changes have been fully reverted. The source-list `embedded_chunks` fix is reviewed separately in #1354. This PR includes only batching, regression tests, and an Unreleased → Fixed changelog entry.

## Related Issue

N/A (small fix; scope narrowed following review).

## Type of Change

- [x] Bug fix
- [x] Performance improvement

## How Has This Been Tested?

- `python -m pytest tests/test_repository_insert.py tests/test_repository_config.py -q`
  - **12 passed in 0.56s**.
  - 120 records produce exactly three INSERTs (50/50/20), with one connection entry and exit, preserving all returned records in order.
  - Covers empty input, invalid batch sizes, exception/string errors, stopping after a failed batch, connection cleanup, and continuation after an ignored duplicate batch.
- Ran the new tests against the previous implementation first: **4 failed, 3 passed**, confirming they catch the repeated connection setup.
- `ruff check open_notebook/database/repository.py tests/test_repository_insert.py`: **All checks passed!**
- `ruff format --check open_notebook/database/repository.py tests/test_repository_insert.py`: **2 files already formatted**.
- `git diff --check`: passed.

These are focused unit regressions; the large-document embedding workflow was not rerun for this revision.

## Design Alignment

Keeps the async database helper minimal and avoids repeated sign-in/close cycles for every batch.

## Checklist

- [x] Reviewed the diff and kept the change scoped
- [x] Added and ran regression tests
- [x] Ran linting and formatting checks
- [x] Added an Unreleased → Fixed changelog entry
- [x] Read the contribution guidelines and vision

No schema migration or breaking API change.

### DIFF
--- a/api/routers/sources.py
+++ b/api/routers/sources.py
@@ -308,6 +308,7 @@ async def get_sources(
             string::lowercase(title OR '') AS title_sort,
             ({SOURCE_TYPE_EXPRESSION}) AS type,
             (SELECT VALUE count() FROM source_insight WHERE source = $parent.id GROUP ALL)[0].count OR 0 AS insights_count,
+            (SELECT VALUE count() FROM source_embedding WHERE source = $parent.id GROUP ALL)[0].count OR 0 AS embedded_chunks,
             (SELECT VALUE id FROM source_embedding WHERE source = $parent.id LIMIT 1) != [] AS embedded
             FROM {from_clause}
             {order_clause}
@@ -360,7 +361,7 @@ async def get_sources(
                     if row.get("asset")
                     else None,
                     embedded=row.get("embedded", False),
-                    embedded_chunks=0,  # Not needed in list view
+                    embedded_chunks=row.get("embedded_chunks", 0),
                     insights_count=row.get("insights_count", 0),
                     created=str(row["created"]),
                     updated=str(row["updated"]),
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -31,7 +31,7 @@ services:
     environment:
       # REQUIRED: Change this to your own secret string
       # This encrypts your API keys in the database
-      - OPEN_NOTEBOOK_ENCRYPTION_KEY=change-me-to-a-secret-string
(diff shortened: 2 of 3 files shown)


lfnovo (changes requested):
Thanks for tracking this down. The batching in `repo_insert` is a legitimate fix: the only caller (`commands/embedding_commands.py`) can pass thousands of records with 3072-dim vectors in one INSERT, and that plausibly exceeds the websocket frame limit behind the `ConnectionResetError`. Three changes before it can merge:

1. **Revert the `docker-compose.yml` change.** That file is what users download for the image-only install (`docs/1-INSTALLATION/docker-compose.md`). Bind-mounting `./open_notebook`, `./commands` and `./api` breaks the shipped image for anyone without a source checkout. If you want those mounts for development, they belong in a separate `docker-compose.dev.yml` override, in its own PR. The `${OPEN_NOTEBOOK_ENCRYPTION_KEY:-...}` interpolation is fine on its own but also unrelated to this fix.
2. **Open the connection once, outside the batch loop.** As written, `db_connection()` is entered per batch, so a 6000-record insert does ~125 signin/close cycles. Hoist it and reuse the connection for all batches. This file is slated to move upstream to surreal-basics, so keeping it minimal and clean matters.
3. **Split the `embedded_chunks` count fix in `api/routers/sources.py` into its own PR.** It looks correct, but it's unrelated to the insert failure and should be reviewed and changelogged separately.

A small regression test for the batching (e.g. 120 records → 3 INSERTs with one connection) and a CHANGELOG line under Unreleased → Fixed would round it out.

PR #1304, opened 2026-09-03
fix: surface missing model and 422 errors in transformation playground

### LINKED ISSUE #1253: [Bug]: Running a Transformation without selecting a Model fails silently (HTTP 422, no UI feedback)

Go to Transformations > Playground; leave "Modelo" ("Model") on its default empty/unselected state ("Selecione um modelo"); type any text into "Texto de Entrada"; click "Transformação Executiva" ("Run Transformation").

### How did it break?

Expected: either the "Run Transformation" button should be disabled until a Model is chosen, or the user should see a clear inline message like "Please select a model." Actual: the button is clickable, the request is sent to the backend, the backend correctly rejects it with HTTP 422, but the frontend does not surface any toast, banner, or inline error — the page just sits there with no feedback at all, making it look like the app silently did nothing.

### Logs or Screenshots

Browser console: Error sending message: AxiosError: Request failed with status code 422 (at .../065evq5sy~pa1.js -> eq -> XMLHttpRequest.b -> tg.request -> Object.sendMessage). No toast/snackbar is rendered in the UI when this fires.

### Open Notebook Version

v1-latest (Docker)

### Environment

- OS: macOS (Docker via OrbStack) - Browser: Google Chrome - Deployment: docker compose, image lfnovo/open_notebook:v1-latest

### Additional Context

_No response_

### Contribution

Alphaxiaoteng:
## Summary
- Fixes #1253: the Transformations playground could fail silently when required fields were missing or when the API returned HTTP 422.
- Show destructive toasts for missing transformation/model/input, catch `mutateAsync` rejections so API errors stay handled by the existing mutation `onError` toast, and map FastAPI validation `detail` arrays to readable toast text.

## Test plan
- [x] `npx vitest run src/lib/utils/error-handler.test.ts` — 2 passed (validation array + string detail)
- [ ] Manual: leave Model empty / trigger 422 and confirm a toast appears
- [ ] CI

AI assistance was used to draft this fix; I reviewed and verified the change.

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1304?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/app/(dashboard)/transformations/components/TransformationPlayground.tsx
+++ b/frontend/src/app/(dashboard)/transformations/components/TransformationPlayground.tsx
@@ -10,6 +10,7 @@ import { ScrollArea } from '@/components/ui/scroll-area'
 import { Play, Loader2 } from 'lucide-react'
 import { Transformation } from '@/lib/types/transformations'
 import { useExecuteTransformation } from '@/lib/hooks/use-transformations'
+import { useToast } from '@/lib/hooks/use-toast'
 import { ModelSelector } from '@/components/common/ModelSelector'
 import { useTranslation } from '@/lib/hooks/use-translation'
 import ReactMarkdown from 'react-markdown'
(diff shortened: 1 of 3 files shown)


lfnovo (changes requested):
Thanks for tackling #1253. The 422 half is right: wrapping `mutateAsync` in try/catch and teaching `formatApiError` to read FastAPI's array-shaped `detail` fixes the blank toast.

The missing-field half can't fire, though. `canExecute` in `TransformationPlayground.tsx` is unchanged and the Execute button is still `disabled={!canExecute}`, so the new toasts for a missing transformation/model/input are unreachable from the UI (cubic flagged the same). Two ways to resolve it:

1. Drop the three toast branches and keep only the 422 fix, or
2. Disable the button only while pending, let `handleExecute`'s guards do the gating, and add a component test that renders the playground with no model selected, clicks Execute, and asserts the toast.

Either is fine with me; (2) matches the issue's intent better. Please also add a one-line CHANGELOG entry under Unreleased → Fixed.

Issue #987, opened 2026-07-01
[Feature]: Openrouter supported for TTS and ASR


kumar-shivang:
### Feature Description

I would like to be able to use openrouter for TTS and ASR as well. Openrouter now host all kinds of models.

### Why would this be helpful?

This help be helpful in finding a diverse set of models on a single place, and might even be cost effective.

### Contribution

- [x] I am a developer and would like to work on implementing this feature (pending maintainer approval)

lfnovo:
OpenRouter is a first-class provider in **esperanto** (our model-abstraction layer), which is where TTS/STT support belongs. I verified OpenRouter now exposes OpenAI-compatible audio endpoints (`/audio/speech` and `/audio/transcriptions`), so this is feasible.

I've filed a fully-specified card upstream: **lfnovo/esperanto#223**. @kumar-shivang — since you offered to implement this, please pick it up there; it has concrete implementation pointers (mirror the existing OpenAI TTS/STT providers pointed at the OpenRouter base URL) and acceptance criteria.

Once esperanto ships the providers, the only change needed here in Open Notebook is adding `text_to_speech` / `speech_to_text` to the OpenRouter modality mapping. Keeping this issue open as the downstream tracker.

PR #1305, opened 2026-09-03
fix: filter OpenRouter discover list by selected model type

### LINKED ISSUE #1274: [Bug]: OpenRouter model discovery never lists embedding models (separate /embeddings/models endpoint not queried)
## Problem

In the Discover Models dialog for an OpenRouter credential, switching Model Type to Embedding still shows only chat models, so no embedding model can be registered from the UI. Reported in the original issue against v1.14.0 (original report preserved below).

## Root cause

- OpenRouter serves chat models at `/api/v1/models` (431 today, no embeddings) and embedding models at a separate `/api/v1/embeddings/models` (33 today). Our discovery only queries the first. Esperanto's OpenRouter embedding provider already uses the second.
- The Model Type dropdown in `DiscoverModelsDialog` is not a filter; it is the type stamped on the selected models at registration. That is fine for providers whose single `/models` catalogue mixes types (OpenAI), but for OpenRouter there is nothing to stamp as embedding.
- `POST /credentials/{id}/discover` drops `model_type` from the response for every provider (`api/routers/credentials.py` builds `DiscoveredModelResponse` without it), so the dialog could not filter even if the backend classified.

Workaround today: type the embedding model id in the search box and use "Add <id>" with type Embedding.

## Fix

Backend (`api/credentials_service.py`, `discover_with_config` openrouter branch):
- Fetch `/api/v1/models` and tag as `language`, and `/api/v1/embeddings/models` and tag as `embedding`.
- Attach `OPENROUTER_AUDIO_MODELS` with the type from their seed entry instead of untyped.

Backend (`api/routers/credentials.py`):
- Pass `model_type` through in `DiscoveredModelResponse` (this also unbreaks Ollama, which already classifies).

Frontend (`frontend/src/components/settings/DiscoverModelsDialog.tsx`):
- When an item has `model_type`, show it only under that type; items without a type stay visible under every type.
- Base `showCustomOption` on the filtered list, and prune `selectedModels` when the type changes.

Tests:
- `tests/test_credentials_api.py`: `discover_with_config("openrouter")` with both endpoints mocked returns typed entries.
- New `DiscoverModelsDialog.test.tsx`: switching to Embedding hides chat models and keeps the custom option.

## Acceptance criteria

- With an OpenRouter credential, selecting Embedding in Discover Models lists the models from `/embeddings/models` and registering one works.
- Language, TTS and STT lists are unchanged for OpenRouter.
- Ollama discovery now returns `model_type` in the API response.
- Other providers' discovery behaviour is unchanged (items without type still appear under every type).

## Related

PR #1305 attempted this on the legacy `GET /models/discover/{provider}` path, which the UI does not call.

---

<details>
<summary>Original report</summary>

1. Create Open Notebook Server via docker compose
2. Go to Models -> Add Openrouter API Config
3. Test -> OK
4. Models -> Default Selection: Language Model Type -> 417 models available -> GOOD
5. Clicking Model Type Dropdown -> Selecting anything other than Language -> Model list still only shows Lanague models -> No embedding model possible to select -> Cant use Open Notebook

See screenshot: Type switched to Embedding, options below are still language models

Image

Hint: registering the models directly via `POST /api/credentials/{credential_id}/register-models` worked and they became visible in the UI. Just the dropdown didn't work.

### Environment

Open Notebook 1.14.0 (v1-latest Docker), Fedora Server

</details>

Alphaxiaoteng:
## Summary
- Fixes #1274: switching Model Type in Discover Models still showed the full chat catalog because OpenRouter discovery forced every id to `language` and the UI did not filter by `model_type`.
- Classify OpenRouter models from `architecture.output_modalities` / id substrings, and filter (plus prune) the dialog list by the selected modality.

## Test plan
- [x] Standalone `_classify_openrouter` checks for language / embedding / whisper / voice
- [x] Updated `test_openrouter_quirk_language_and_description` expectations
- [ ] CI / frontend manual: Discover Models → switch to Embedding → list shows embedding-classified ids only

AI assistance was used to draft this fix; I reviewed and verified the change.

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1305?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/components/settings/DiscoverModelsDialog.tsx
+++ b/frontend/src/components/settings/DiscoverModelsDialog.tsx
@@ -90,12 +90,30 @@ export function DiscoverModelsDialog({
     setCustomModelSelected(false)
   }, [searchQuery])
 
-  // Filter discovered models by search query
+  // Filter discovered models by selected modality first, then search query.
+  // Without the type filter, switching Model Type left the full chat catalog
+  // visible (#1274) even though registration would stamp a different type.
   const filteredModels = useMemo(() => {
-    if (!searchQuery.trim()) return discoveredModels
+    const byType = discoveredModels.filter(
+      (m) => !m.model_type || m.model_type === selectedType
+    )
+    if (!searchQuery.trim()) return byType
     const q = searchQuery.toLowerCase()
-    return discoveredModels.filter(m => m.name.toLowerCase().includes(q))
-  }, [discoveredModels, searchQuery])
+    return byType.filter((m) => m.name.toLowerCase().includes(q))
+  }, [discoveredModels, searchQuery, selectedType])
+
+  // Drop selections that no longer belong to the active model type.
+  useEffect(() => {
+    setSelectedModels((prev) => {
+      const allowed = new Set(
+        discoveredModels
+          .filter((m) => !m.model_type || m.model_type === selectedType)
+          .map((m) => m.name)
+      )
+      const next = new Set([...prev].filter((name) => allowed.has(name)))
+      return next.size === prev.size ? prev : next
+    })
+  }, [discoveredModels, selectedType])
 
   // Show custom model option when search doesn't exactly match any discovered model
   const showCustomOption = useMemo(() => {
--- a/open_notebook/ai/model_discovery.py
+++ b/open_notebook/ai/model_discovery.py
@@ -250,11 +250,39 @@ class ProviderDiscoverySpec:
     description: Optional[Callable[[dict], Optional[str]]] = None
 
 
+def _classify_openrouter(model: dict) -> str:
+    """Classify OpenRouter catalog entries using architecture metadata + name.
+
+    OpenRouter's /models payload is dominated by chat models, but embedding
+    (and some audio) ids are present. Blindly tagging everything as ``language``
+    made the discover UI show the same chat list after switching Model Type
+    (see #1274). Prefer ``architecture.output_modalities`` when present, then
+    fall back to id substrings, then language.
+    """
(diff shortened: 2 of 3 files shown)


lfnovo (changes requested):
Thanks for picking up #1274. Unfortunately the change lands on a code path the UI doesn't use, so the dialog still shows the full catalog.

The Discover Models dialog calls `POST /credentials/{id}/discover`, served by `discover_with_config()` in `api/credentials_service.py`. That function only sets `model_type` in the Ollama branch; the OpenRouter branch returns names with no type. `_classify_openrouter()` was added to `open_notebook/ai/model_discovery.py`, which backs `GET /models/discover/{provider}`, and no frontend component calls that endpoint. As a result `m.model_type` is always `undefined` in `DiscoverModelsDialog`, and the new filter `!m.model_type || m.model_type === selectedType` is always true.

To fix #1274 for real:
- Reuse the classification inside `discover_with_config`'s OpenRouter branch so each discovered entry carries `model_type`.
- Make sure the seeded `OPENROUTER_AUDIO_MODELS` entries (#987) carry their known TTS/STT type, otherwise a correct filter would hide them.
- Add a test against the credential-based discover path (the current test only exercises `discover_openai_compatible_provider`) and a small frontend test for the dialog filter.

PR #1307, opened 2026-09-03
fix(transformations): make insight max_tokens configurable

### LINKED ISSUE #1273: [Bug] Source insights (transformations) silently truncated at Esperanto default max_tokens=850 — no way to override for openai_compatible / non-reasoning models
**Environment:** open-notebook 1.14.0 (`v1-latest-single`), esperanto 2.25.1, provider `openai_compatible` (vLLM, served model `Qwen3.8`, 256K context window).

## Problem

Source insights (transformations) are silently cut off mid-sentence / mid-word. The truncation correlates directly with the transcript length — long sources get cut, short ones complete cleanly:

| full_text (chars) | insight (chars) | result |
|---|---|---|
| 13431 | 1228 / 2002 | truncated (e.g. ends `…Eine form`) |
| 16581 | 11333 | truncated (mid list item) |
| 325 | 3832 | complete (ends with `.`) |
| 715 | 3021 | complete |
| 451 | 2449 | complete |

## Root cause

Transformations provision the language model via `provision_langchain_model()` (`open_notebook/ai/provision.py`) -> `model.to_langchain()`. No explicit `max_tokens` is passed anywhere in this path.

Esperanto's `LanguageModel` base default is `max_tokens = 850` (`esperanto/providers/llm/base.py:19`). The OpenAI provider only *skips* this default cap for reasoning models (`_is_reasoning_model()` = name starts with `o1/o3/o4/gpt-5`, `esperanto/providers/llm/openai.py:316`). An `openai_compatible` model such as `Qwen3.8` is **not** detected as reasoning, so every transformation request is sent with `max_tokens=850`.

There is no way to override this from Open Notebook:
- `ModelCreate` / `ModelResponse` have no `max_tokens` field.
- `Credential.to_esperanto_config()` only emits known keys (`api_key`, `base_url`, `endpoint*`, `num_ctx`, ...) — arbitrary keys like `max_tokens` are dropped.
- No relevant env var / setting exists.

**Verified live** in the container:
```python
m = await model_manager.get_default_model("transformation")   # OpenAICompatibleLanguageModel "Qwen3.8"
m.max_tokens                 # -> 850
m.to_langchain().max_tokens  # -> 850  (ChatOpenAI)
```

The backend model itself is not the limit — a direct request to the same endpoint with `max_tokens=3000` returns a complete 2852-token answer (`finish_reason=stop`). So the cap is imposed solely by Open Notebook/Esperanto.

## Suggested fix

Analogous to the Ask fix in #1241 (`OPEN_NOTEBOOK_ASK_MAX_TOKENS`): make the transformation output budget configurable — e.g. an `OPEN_NOTEBOOK_TRANSFORMATION_MAX_TOKENS` env var and/or a per-model `max_tokens` in the model/credential config — and/or stop silently falling back to Esperanto's 850 default for non-reasoning `openai_compatible` models.

## Related

#1221, #1247, #1241 (Ask/Q&A truncation), #1200 (provider-declared tunable params), #947 (model & provider advanced config).

Alphaxiaoteng:
## Summary
- Adds `OPEN_NOTEBOOK_TRANSFORM_MAX_TOKENS` (default `8192`) so source insights / transformations are not stuck on Esperanto’s silent 850-token default for non-reasoning `openai_compatible` models.
- Invalid or non-positive values fall back to `8192` with a warning.
- Documents the env var and covers it with focused regressions.

Fixes #1273

## Test plan
- [x] `pytest tests/test_transform_token_budget.py` (8 passed)

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1307?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -87,3 +87,6 @@ SURREAL_DATABASE=open_notebook
 
 # For more configuration options, see:
 # https://github.com/lfnovo/open-notebook/blob/main/docs/5-CONFIGURATION/environment-reference.md
+
+# Max output tokens for source insights / transformations (default: 8192)
+# OPEN_NOTEBOOK_TRANSFORM_MAX_TOKENS=8192
--- a/docs/5-CONFIGURATION/environment-reference.md
+++ b/docs/5-CONFIGURATION/environment-reference.md
@@ -330,3 +330,9 @@ If you have these variables configured from a previous installation, click the *
 | `AZURE_OPENAI_API_KEY_EMBEDDING` | Azure OpenAI | Configure per-service in credential |
 | `AZURE_OPENAI_ENDPOINT_EMBEDDING` | Azure OpenAI | Configure per-service in credential |
 | `AZURE_OPENAI_API_VERSION_EMBEDDING` | Azure OpenAI | Configure per-service in credential |
+
+### `OPEN_NOTEBOOK_TRANSFORM_MAX_TOKENS`
+
+Maximum output tokens for source insights / transformations. Default: `8192`.
+Invalid or non-positive values fall back to `8192`.
+
--- a/open_notebook/graphs/transformation.py
+++ b/open_notebook/graphs/transformation.py
@@ -1,7 +1,11 @@
+import os
+from functools import cache
+
 from ai_prompter import Prompter
 from langchain_core.messages import HumanMessage, SystemMessage
 from langchain_core.runnables import RunnableConfig
 from langgraph.graph import END, START, StateGraph
+from loguru import logger
 from typing_extensions import TypedDict
 
 from open_notebook.ai.provision import provision_langchain_model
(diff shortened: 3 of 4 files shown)


lfnovo:
Thanks for this. The change itself is sound, and I'm parking it deliberately rather than for quality reasons: we're consolidating how max_tokens (and the other model tunables) get configured under #947, with #830 moving them into Settings instead of one environment variable per workflow. I'd rather not add another env var now and deprecate it a release later. I'll revisit this PR once #830 lands; if it ends up being the natural place for this knob, we'll fold it in with credit.

lfnovo:
Closing without merge, for the same reason as #1247: model tunables are being consolidated per model under #947 / #830 rather than one environment variable per workflow. Also worth knowing before more work goes into this path: the issue's root cause doesn't hold — transformations already pass `max_tokens=8192`, and it reaches the model (details on #1273). The variable would let the reporter raise the number, but the underlying problem looks like thinking tokens eating the budget, which needs different handling. Thanks for the clean implementation; the validation helper is a good shape for whatever #830 ends up doing.

PR #1334, opened 2026-09-05
fix(podcasts): restore the language instruction and stop the prompts handing the model a copyable skeleton

alefbt:
## Description

First of the two PRs #1239 was split into, per @lfnovo's review. This one is the low-risk half: the prompt templates and the failure hints. The voice pre-flight stays in #1239 for its own review.

**1. `EpisodeProfile.language` is inert.** `prompts/podcast/{outline,transcript}.jinja` shadow podcast-creator's bundled templates — the library resolves `Path.cwd()/prompts/podcast/<name>.jinja` before its own package resources, and this app only calls `configure("speakers_config"/"episode_config")` — and the app's copies never referenced `{{ language }}`. The value is plumbed correctly end to end (`graph.py:100` resolves it, `graph.py:133` maps the code to a name, `nodes.py` passes it), so a profile set to `he-IL` produced an English outline and English segment titles. The field looked supported and did nothing.

Restored in both templates. A regression test now fails when a variable the bundled template uses is missing from the app's copy — the drift that lost it.

**2. The prompts handed the model a skeleton it could copy.** Both templates showed a fill-in JSON skeleton (`"speaker": "[Actual Speaker Name]"` plus a bare `...`) inside a ` ```json ` fence, while also instructing the model not to use fences. Gemini returned the skeleton verbatim, which failed podcast-creator's speaker-name validation and aborted the episode at segment 2/6 — discarding the outline and the segment already generated:

```
Value error, Invalid speaker name '...'. Must be one of: Marcus Thompson, Elena Vasquez, Johny Bing
```

The example now carries the episode's real speaker names (via `tojson`, so a name containing a quote can't break it) and written-out sample values, so a verbatim copy is *valid* output. Placeholders and truncation are banned explicitly, the contradictory fence and the second copyable example are gone, and the example is labelled a two-entry excerpt restating the required `turns` / `num_segments` count so it can't read as a finished answer.

Per your last note, **no sample is rendered when a `language` is set** — any hard-coded sample is English, and a copied English line in a Hebrew episode goes straight to TTS. The structure is stated in prose there, and `format_instructions` still carries the schema.

**3. Failure hints.** `explain_generation_failure()` maps a failure to the hint that fits. The GPT-5 extended-thinking note was the only one, keyed on `Invalid json output` / `Expecting value`, so `Invalid speaker name` got nothing and a truncated Gemini response was told to switch to `gpt-4o`.

## Related Issue

Part of #1238. Split from #1239.

## Type of Change

- [x] Documentation update

## How Has This Been Tested?

- [x] Added new unit tests

**Test Details:** 768 tests pass, `ruff check .` and `mypy` clean. New: `tests/test_podcast_prompt_templates.py` (renders both templates: language block present/absent, example parsed with `json.loads` for speaker fidelity and escaping, no placeholder skeletons, no sample at all when a language is set, and the bundled-vs-app variable drift check) and `tests/test_podcast_error_hints.py`.

Verified the tests are genuine regressions against pre-fix templates: `[Actual Speaker Name]` present, `{"segments": [...]}` present, no language block with `language="Hebrew"`, and `language` reported missing from both templates.

Not smoke-tested end to end from this branch — the end-to-end Hebrew run (6 segments, 42 lines, MP3, 474 s) was on the reporting deployment with these templates bind-mounted.

## Design Alignment

**Explanation:** The pipeline silently assumed English output; a non-English deployment got a podcast in the wrong language with no signal. No new configuration or endpoints — the prompts just stop contradicting themselves.

### Code Quality
- [x] I have commented my code, particularly in hard-to-understand areas

### Testing
- [x] I have added tests that prove my fix is effective
- [x] New and existing unit tests pass locally with my changes
- [x] I ran linting: `ruff check . --fix`
- [x] I ran type checking: `uv run python -m mypy .`

### Documentation
- [x] I have updated the relevant documentation in `/docs` — `docs/7-DEVELOPMENT/podcasts.md` gains a "Prompt templates shadow podcast-creator's" section: the resolution order, which variables each template receives, and why no sample is rendered for a non-English run

### Database Changes
- [ ] No schema changes

### Breaking Changes

## Additional Context

The deeper question from #1238 stands and is deliberately not answered here: these templates have diverged from the library's in both directions — they *gained* the solo-speaker branches and the `<think>`-tag rules and *lost* the language block. Either the app-only features get upstreamed into podcast-creator and `prompts/podcast/` is deleted, or the fork is documented as intentional. The drift test makes the next divergence fail loudly in the meantime.

- [x] I have read [CONTRIBUTING.md](https://github.com/lfnovo/open-notebook/blob/main/docs/7-DEVELOPMENT/contributing.md) and [VISION.md](https://github.com/lfnovo/open-notebook/blob/main/VISION.md)
- [x] This PR addresses a reproducible bug (#1238), reviewed and split at maintainer request
- [x] My PR title follows conventional commits format

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1334?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


alefbt:
All four cubic findings were valid; fixed in [`9228bb9`](https://github.com/alefbt/open-notebook/commit/9228bb9), and the branch is rebased onto current `main`.

**The first one is worse than reported, and it's the important one.** `except ValueError: raise` sits ahead of the handler that attaches hints, and both `OutputParserException` and `json.JSONDecodeError` are `ValueError` subclasses — I verified both. So *every* parser failure skipped the mapper: the placeholder speaker name, `Invalid json output`, and truncated responses. That means the **pre-existing GPT-5 extended-thinking note has been dead since it was added**, for exactly the failure it describes. The `ValueError` branch now attaches the hint and re-raises a `ValueError`, so the command layer's permanent-failure contract (`stop_on=[ValueError]`) is unchanged — a new test asserts both the hint and the preserved type, and a second asserts an unrelated `ValueError` still passes through untouched.

**`Requested entity was not found` no longer blames the voice.** Right that Google returns it for any missing resource — a bad model id on the outline or transcript call included. The hint now names both candidates and says which is likelier when audio had already started; the voice-specific hint stays keyed on `Voice name … is not supported`, which does name the voice.

**Solo-speaker contradiction** — fixed; the excerpt wording follows `speakers|length`, so a one-speaker profile reads "one-entry EXCERPT … rather than the one shown".

**`not` counted as a template variable** — fixed with a keyword exclusion set. Harmless today, but it would have failed the drift check the moment one template negated a test the other didn't, which is precisely the check's job.

772 tests pass, ruff and mypy clean; the two new hint assertions fail against `6fb1510`.

PR #1339, opened 2026-09-07
fix(search): resolve insight text hits to parent sources

### LINKED ISSUE #1312: [Bug]: search result parent_id for insights differs between text and vector search

I called POST /api/search twice for the same content with two different type values (text then vector), in order to map each hit back to its source via parent_id (e.g. to filter results down to a single notebook using the source→notebook relationship).

### How did it break?

The parent_id field on insight results (source_insight:…) is inconsistent between the two search types:

type: "vector" → parent_id correctly points to the parent source (source:…) ✅
type: "text" → parent_id points to the insight itself (source_insight:…), not to its source ❌

Regular source-content hits (source:…) are consistent in both modes (id == parent_id); only insights differ.

Impact: any client that resolves a hit back to its source through parent_id (e.g. to scope/attribute results to a notebook) works with vector but silently breaks with text — insight hits can't be matched to a source, so they get dropped or mis-attributed.

Expected: parent_id on insight results should always point to the parent source (source:…), regardless of type, matching the vector behavior.

### Logs or Screenshots

// type: "text"  → parent_id = the insight itself (WRONG)
{ "id": "source_insight:bot5t9zkquqrtx5c4uf6",
  "parent_id": "source_insight:bot5t9zkquqrtx5c4uf6",
  "relevance": 9.04, "title": "Key Insights SEO - …" }

// type: "vector" → parent_id = the parent source (CORRECT)
{ "id": "source_insight:bot5t9zkquqrtx5c4uf6",
  "parent_id": "source:jx5rfiml38tcfeqptnoz",
  "similarity": 0.58, "title": "Key Insights SEO - …" }

### Open Notebook Version

v1-latest (Docker)

### Environment

_No response_

### Additional Context

_No response_

### Contribution

hydraxman:
## Description

Text-search insight hits currently return the insight itself as `parent_id`. Clients that use this field to attribute a hit to its source silently lose those results; vector search already returns the parent source correctly.

Migration 25 changes the insight projection to `source.id as parent_id`, while retaining the insight's own `id`. It replaces only `fn::text_search`; vector search, notebook filtering, source/note results, ranking and limits are unchanged. The down migration restores the previous text function. Migration 24 remains untouched, per ADR-006.

## Related Issue

Fixes #1312

## Type of Change

## How Has This Been Tested?

- [x] Added new unit tests
- [x] Existing tests pass

The regression uses the existing Python `surrealdb` dependency's embedded in-memory engine, not mocked query results. It applies the search schema and registered search migrations; it does not start the unrelated podcast services or call a model provider.

Before the fix, all three scope cases failed with `parent_id == source_insight:insight` rather than `source:parent`.

After the fix:
- `uv run pytest tests/test_text_search_parent.py tests/test_search_api.py -q --tb=short` — 38 passed.
- `uv run pytest tests/ -q --tb=short` — 721 passed, 2 dependency deprecation warnings.
- `uv run ruff check .` — passed.
- `uv run python -m mypy .` — no issues in 135 source files.
- `git diff --check` — passed.

Coverage includes global, empty and selected-notebook scopes; exclusion of another notebook; source-only/note-only filters; source and note identity controls; empty results; and an up/down/up cycle that checks the vector function definition is unchanged.

## Design Alignment

- [x] API-First Architecture

Correct the database projection at its source rather than add an API-specific workaround. This changes no stored records or search signatures.

## Checklist

- [x] Performed a self-review and kept unrelated changes out.
- [x] Added regression coverage and ran existing Python tests, lint and type checking.
- [x] Added and registered both up and down migrations; tested the migration cycle locally.
- [x] Read the contribution guide, backend rules and migration policy.
- [x] This is a small, obvious bug correction under the contribution guide's assignment exemption; the new migration files preserve already-shipped migrations.

No frontend or provider behavior was changed; frontend checks were not run locally.

<a href="https://cubic.dev/pr/lfnovo/open-notebook/pull/1339?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lfnovo (approved):
The diff is exactly as narrow as it claims. `diff 24.surrealql 25.surrealql` shows one functional change — `id as parent_id` → `source.id as parent_id` in the `$source_insight_search` branch — and `25_down.surrealql` is byte-for-byte identical to migration 24's `fn::text_search`, so the down really restores the previous state rather than an approximate reconstruction. Notebook scoping, `fn::vector_search`, ranking, grouping and limits are untouched.

Correcting the projection in the database is the right layer: `api/routers/search.py` passes rows straight through, so `parent_id` is a SQL contract, not an API one. ADR-006 is respected — new migration with a registered `_down`, migration 24 left alone.

The tests earn their keep. They run the real migrations against the embedded `mem://` engine instead of mocking query results, and cover all three scope cases, cross-notebook exclusion, the source-only/note-only filters, and an up/down/up cycle that asserts the vector function definition is unchanged.

Two notes, neither blocking:

- `frontend/src/app/(dashboard)/search/page.tsx:475` derives the modal type from `parent_id`. With this fix, a text hit on an insight will open the source modal rather than the insight modal — consistent with vector search, which already behaved this way, but a change for text. The frontend should switch to `result.id`, which is now unambiguous in both modes. Filed as a follow-up.
- `tests/test_text_search_parent.py:42` — `up_migrations[23:]` is an open-ended slice, so a future migration that depends on podcast-creator tables would break this test by accident. `[23:25]` would state the intent more precisely. Micro-detail, fine as is.

Thanks for the careful work on this one.

